Issue #24, opened 2017-03-18
Severe quantization issues with Canon CR2 files.


KelSolaar:
Hi,

I performed some initial tests using [some of my Canon 5D Mark II CR2 files](https://github.com/colour-science/colour-hdri-examples-dataset/tree/master/frobisher_001) and experienced some severe quantization issues:

![image](https://cloud.githubusercontent.com/assets/99779/24067669/b89515aa-0be4-11e7-9bce-f38e8206db16.png)

- [IMG_2601_colour-hdri.exr (203MB)](https://drive.google.com/open?id=0B_IQZQdc4Vy8Y05PTVo2Z2VKLVU): It uses [Colour - HDRI](https://github.com/colour-science/colour-hdri), for this particular image, I'm not using LibRaw but the following conversion path: CR2 to DNG conversion with Adobe DNG Converter, hence the low quality demosaicing) ---> DNG to Linear Tiff with DCRaw ---> Python Sauce ---> sRGB EXR
- [IMG_2601_aces.exr (121MB)](https://drive.google.com/open?id=0B_IQZQdc4Vy8YjI1NnBzemRNSTg): `rawtoaces --mat-method 1 --wb-method 1 IMG_2601.CR2`

Cheers,

Thomas

PR #25, opened 2017-03-21
Fixes a bug related to CR2 RAW

miaoqi:
* reformats and cleans code in updated source files
* changes the camera model from “5d” to “eos 5d max ii” on the second line of “data/camera/Canon_5D_380_780_5”
* updates “loadCameraSpst ()” function of Idt class to determine if the given spectral sensitivity data is external
* uses “uchar” to “ushort” for image “data” array type-conversion after calling “dcraw_process ()” for further operations that are based on floating-point values
* adds additional condition in “prepareIDT ()” and “prepareWB()” to make sure the spectral sensivity data is eternal or already included
* solve #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ If there are errors with regard to version mis-match of *Eigen* library, please 
 	$ sudo make install
 	```
 
-	The default process will install `librawtoacesIDT.dylib` to `/usr/local/lib`, a few header files to `/usr/local/include/rawtoaces` and a number of data files into `/usr/local/include/rawtoaces`
+	The default process will install `librawtoacesIDT.dylib` to `/usr/local/lib`, a few header files to `/usr/local/include/rawtoaces` and a number of data files into `/usr/local/include/rawtoaces/data`
 	
 ## Usage
 


data/camera/Canon_5D_380_780_5
```diff
@@ -1,5 +1,5 @@
 canon
-5d
+eos 5d mark ii
```
lchen198, line 2: Will you consider changing this file into a json or other self-describing data types

lib/mathOps.h
```diff
@@ -271,7 +271,8 @@ void minusVector(vector<T>& vct, T sub)
 }
 
 template <typename T>
-vector<T> mulVectorElement(const vector<T>& vct1, const vector<T>& vct2)
```
lchen198, line 274: Do you have test for these low-level functions? I guess you must have some constraint that you can not use alternative well-tested libraries

src/acesrender.cpp
```diff
@@ -156,8 +157,9 @@ int AcesRender::readCameraSenPath( const char * cameraSenPath,
                 string fn( *file );
 
                 readC = _idt->loadCameraSpst( fn,
-                                             static_cast <const char *> (P.make),
-                                             static_cast <const char *> (P.model) );
+                                              static_cast <const char *> (P.make),
```
lchen198, line 160: Do you have a code formating tool that can be used by everyone on the team?

lib/mathOps.h
miaoqi: That is a good point. I have though about using eigen library.

src/acesrender.cpp
miaoqi: Can you recommend some tools? Thanks!

lib/mathOps.h
miaoqi: Are you aware of any good math libraries that are available on various platforms including these old Linux systems?

Issue #26, opened 2017-03-22
use JSON file for data


miaoqi:
As @lchen198 suggest, there should be a common and standard format to host the data such as camera spectral sensitivity data.  "JSON" is one choice.

KelSolaar:
Just pushing that from Slack for further discussion regarding #31:

Almost complete for me so in the light of doing that there might be some small changes to do, essentially attribute names, taking the stuff I was saying before:

- Both `model` and `catalog_number` are in use, any reasons for using both?
- Would like to rename `api_version` to `schema_version`.
- Tempted to rename `spectral_data` to `body`.
- Would like to rename `data` to `dataset` as we are dealing with multiple data set.
- `document_creation_date` should respect RFC 3339.

PR #47, opened 2017-05-24
Update "JSON" dataset dates to be compatible with "RFC 3339".

KelSolaar:
References ampas/rawtoaces#26.


KelSolaar:
I noticed that some extra end-line spaces got culled by my commit. I can revert that if needed.

miaoqi:
@aforsythe @KelSolaar Looks good! Thanks!

PR #59, opened 2017-06-02
Feature/fixhiddenfiles

### LINKED ISSUE #58: .DS_store files
It appears that every time I reinstall to test updates, cmake is moving .DS_store files from `../rawtoaces/data/camera` into `/usr/local/include/rawtoaces/data/camera`

In the install configuration I'm seeing this line:
`-- Installing: /usr/local/include/rawtoaces/data/camera/.DS_Store`

which, when I try to run rawtoaces, gives me this error

```
/usr/local/include/rawtoaces/data/camera/.DS_Store(1): expected value
Please double check the Camera Sensitivity data (e.g. the increment should be uniform from 380nm to 780nm).
```

Removing the files fixes the issue, but I can see this being a potential problem for users who use Finder to add specsens or other data into rawtoaces.

Additionally, .DS_store files should likely be added to the .gitignore

miaoqi:
- fixes the bug that will try to read unneeded file such as ".DS_store" during data loading. 
- closes #58

### DIFF
--- a/src/acesrender.cpp
+++ b/src/acesrender.cpp
@@ -144,7 +144,7 @@ int AcesRender::readCameraSenPath( const char * cameraSenPath,
 
             for ( vector<string>::iterator file = cFiles.begin( ); file != cFiles.end( ); ++file ) {
                 string fn( *file );
-                
+                if ( fn.find(".json") == std::string::npos ) continue;
                 readC = _idt->loadCameraSpst( fn,
                                               static_cast <const char *> (P.make),
                                               static_cast <const char *> (P.model),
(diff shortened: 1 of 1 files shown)


miaoqi:
You are right that this commit is indirectly connected to hidden file fix. It is directly related to data loading. I forgot to create another branch to address this specific commit.

src/acesrender.cpp
```diff
@@ -144,7 +144,7 @@ int AcesRender::readCameraSenPath( const char * cameraSenPath,
 
             for ( vector<string>::iterator file = cFiles.begin( ); file != cFiles.end( ); ++file ) {
                 string fn( *file );
-                
+                if ( fn.find(".json") == std::string::npos ) continue;
```
aforsythe, line 147: Just a thought, but is it better to only read files with a .json extension into the cFiles variable instead of testing each file in cFiles to see if it's a .json?

aforsythe:
I've removed the unrelated commit 4b94ce1b1631b8bb513526e7b831dc053ca589db from #59.  I placed a branch on my fork with it here fab0c0dc65f96357836d71abb07d3141e38fffca

src/acesrender.cpp
miaoqi: The common method of "opendir (path)" will open a path and return a list of files in that path to "cFiles". I may be able to find a method that only returns files of a certain type such as JSON.

aforsythe: I understand.  Don't worry about it.

PR #62, opened 2017-06-02
Fix incorrect link in "README.md" file.

KelSolaar:


KelSolaar:
A ton of links are actually broken, I'll fix them in another PR I think,

PR #78, opened 2017-06-23
Feature/cleanup

miaoqi:
* cleans up and simplifies the code base;
* removes the flag of "--ss-path" and the feature to read from external camera sensitivity file;
* adds new flags such as "--valid-illums" and associated functions 
* fixes a bug related to "half-size"
* changes "make" and "installation" processes
* updates "README.md" to reflect the changes above


src/usage.h
```diff
@@ -324,57 +321,57 @@ int configureSetting ( int argc,
...
+                        lowerCase ( opts.illumType );
                     }
                     else {
-                        fprintf( stderr,"\nError: white balance method 1 requires an illuminant be specified" );
+                        fprintf( stderr, "\nError: white balance method 1 requires an "
+                                         "illuminant be specified\n" );
                         exit(-1);
                     }
                 }
-                if ( opts.use_wb == 3 ) {
```
lchen198, line 346: using const variable or an enum would make it more clear.

miaoqi: Cool! Thanks! Will do it now.

miaoqi: Just fixed it. :)

PR #83, opened 2017-07-05
Feature/day light

miaoqi:
This branch incorporates "daylight" math functions to calculate Daylight Spectral Power Distribution (SPD) dataset instead of reading from external JSON files.


aforsythe:
@miaoqi seems like it's failing with an `assert` but probably should be some error checking and fail with an `cerr` message.

`./rawtoaces --wb-method 1 D6023 /Users/alex/Desktop/testing/raw/AFA_9285.NEF 
Assertion failed: (cct >= 4000 && cct <= 25000), function cctToxy, file /Users/alex/Source/rawtoaces/lib/rta.cpp, line 207.
Abort trap: 6`

Also, it fails when typing in a specific cct ... e.g. d6025.

miaoqi:
I thought daylight is like "D60", "D65", "D250", etc. How can it reach "D6025"?

aforsythe:
D65 is short for Daylight with a CCT of 6504 .. Daylight can be any CCT between 4000K and 25,000K

Issue #97, opened 2017-11-22
libsuitesparseconfig.4.5.5.dylib "Library not loaded" at runtime on OS X


agentirons:
OS X El Capitan 10.11.6

Did a ```$ git pull``` on the most recent version of rawtoaces, ran cmake, make, and make install with no errors, then on running rawtoaces I received this output:
```
dyld: Library not loaded: /usr/local/opt/suite-sparse/lib/libsuitesparseconfig.4.5.5.dylib
  Referenced from: /usr/local/opt/ceres-solver/lib/libceres.1.dylib
  Reason: image not found
Trace/BPT trap: 5
```
The solution was to update my brewed version of ceres-solver, which needed the command ```$ brew reinstall --HEAD ceres-solver```

I can work on a pull request to update the readme tomorrow with mention of this solution in the ceres-solver installation step.

agentirons:
This issue appears to be fixed with the new Homebrew version of rawtoaces.

PR #98, opened 2017-11-22
Updated readme with solution to suite-sparse error

agentirons:
Solution for #97 added to readme.md, also added to the wiki installation page.


agentirons:
The issue this solved has been fixed with the newest version of rawtoaces, so closing this request.

PR #108, opened 2018-04-24
XYZ to ACES matrix

kisakata:
Values of XYZ to ACES matrix defined in XYZ_acesrgb was not correct.
It was using value with D65 to D50 Bradford matrix applied.

Since XYZ to ACES matrix was incorrect,
renderNonDNG() in acesrender.cpp was applying chromatic adaptation twice.
 - D50 to D60 CAT02 matrix in applyCAT()
 - D65 to D60 Bradford matrix that was included in XYZ_acesrgb
I updated XYZ_acesrgb to the value defined in ACES specification,
so that chromatic adaptation will be applied only once.

Since rta.cpp uses XYZ_acesrgb too, I 
I created a variable called XYZ_D65_acesrgb and copied the original value of XYZ_acesrgb,
and updated rta.cpp to use XYZ_D65_acesrgb, so that it still uses same value.

One other thing I noticed is that converted values in XYZ colorspace is not in D50, but in D65.
(LibRaw converts RAW image once to sRGB and then converts to XYZ,
so the values in XYZ are in D65)
I've defined variable called d65, and applyCAT uses this value instead of d50.


lib/define.h
```diff
@@ -174,6 +174,7 @@ const double dmax = numeric_limits<double>::max();
 static const double XYZ_w[3] = {0.952646074569846, 1.0, 1.00882518435159};
 static const double d50[3] = {0.9642, 1.0000, 0.8250};
 static const double d60[3] = {0.952646074569846, 1.0000, 1.00882518435159};
+static const double d65[3] = {0.9547, 1.0000, 1.0883};
```
spatre, line 177: The correct XYZ values for d65 should be [0.95047,1.00000,1.08883]

d65

spatre:
The entire Pull request is valid except for a typo on line 2 in the summary: "It was using value with **D65 to D50** Bradford matrix applied." We've confirmed that the XYZ_acesrgb matrix is in fact using a **D65 to D60** Bradford matrix. Your code changes are being verified by the admin, to be updated to the source master.

kisakata:
Thanks @spatre, for catching the typo.

antond-weta (approved):

PR #118, opened 2020-03-31
PR: Add "Dockerfile".

KelSolaar:
This PR adds a *Dockerfile* to run *rawtoaces* in a container with [Docker](https://www.docker.com/):

Assuming that you have Docker and this PR pulled:

*Building the Container*

Note that the compilation of *ceres-solver* requires a significant amount of memory, well over the 2Go that Docker allocates by default. Thus you will need to increase the memory in Docker preferences: **Preferences --> Resources --> Advanced**, 8Go should be enough.

```sh
docker build -f "Dockerfile" -t rawtoaces:latest "."
```

*Running the Container*

```sh
docker run -it --rm -v $PWD:/tmp -w /tmp rawtoaces:latest rawtoaces IMG_1234.CR2
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM ubuntu:bionic
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends apt-utils dialog 2>&1 \
+    && apt-get -y install curl git  \
+    && apt-get -y install cmake build-essential libilmbase-dev libraw-dev libboost-all-dev libgoogle-glog-dev libatlas-base-dev libeigen3-dev libsuitesparse-dev \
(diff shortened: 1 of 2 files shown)


notwa:
as a bit of an aside, I've tweaked your dockerfile to make it a little more robust, and shrunk the final image by not including developer tools. note, however, that my version isn't suitable as a pull request since I've hardcoded certain commits to checkout (including those that come from a fork (#108)), but I welcome anyone to adapt it.
https://gist.github.com/notwa/89e05a60ea5923114d28302e17bd3295

antond-weta (approved):

antond-weta:
@KelSolaar could you please resolve the conflicts there?

KelSolaar:
Should be good @antond-weta! I will work on the ASWF based container in another PR.

PR #131, opened 2023-11-11
Use "ASWF" container as base.

KelSolaar:
This PR uses the `aswf/ci-vfxall:2022` container as a base rather than *Ubuntu* previously.

Note that the 2023 container did not work when I tried, hence the 2022.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,44 @@
-FROM ubuntu:bionic
+# https://hub.docker.com/r/aswf/ci-vfxall/tags
+FROM aswf/ci-vfxall:2022
 
-ENV DEBIAN_FRONTEND=noninteractive
+# Base Building Dependencies.
+RUN yum install --setopt=tsflags=nodocs -y \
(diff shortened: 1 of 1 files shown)


antond-weta:
@KelSolaar Hey Thomas, on the 2023 container you can "yum install ceres-solver-devel LibRaw-devel".
No need to build them from sources.

KelSolaar:
Hey,

The Centos package Libraw version is too old (0.19) and is missing the `libraw_dng_levels_t. baseline_exposure` [struct member](https://github.com/LibRaw/LibRaw/blob/41506ef73e33c5e0c3b94608cce2481d92003d5d/libraw/libraw_types.h#L251) and the original *rawtoaces* code was developed against Ceres 1.14 but the Centos package is 1.12. I would be inclined to keep it this way if possible.

Cheers,

Thomas

antond-weta:
got it, thanks

antond-weta (approved):

PR #133, opened 2023-11-14
[fix] hardcoded data path

antond-weta:
fix a bug with the path environment variable not being respected and a hardcoded search location always used instead


KelSolaar (approved):
LGTM!

KelSolaar (approved):

PR #135, opened 2023-12-07
[fix] bug in matrix-vector multiplication

antond-weta:
There was a bug in the matrix-vector multiplication where the calculated components were modified in place. So the red component would get calculated, put into the original place, and then the new value would be used when calculating the green and blue instead of the original one.


KelSolaar (approved):
Thanks! LGTM!

PR #146, opened 2025-04-18
add more CI platforms

antond-weta:
Adds VFX2022 and VFX2024 CI runners.
Fixes the build on VFX2022 (Centos 7).
Fixes a few deprecation warnings.
Fixes a potential crash writing out of bounds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,75 @@ on:
     #branches: [ "master" ]
 
 jobs:
+  aswf_old:
+    name: "VFX${{matrix.vfxyear}} ${{matrix.desc}}"
+    strategy:
+      # Set fail-fast to false to ensure that feedback is delivered for all matrix combinations. Consider changing this to true when your workflow is stable.
+      fail-fast: false
+      
+      matrix:
+        include:
+          - desc: gcc11
+            nametag: linux-vfx2022-gcc11
+            os: ubuntu-latest
+            container: aswf/ci-osl:2022-clang11
+            vfxyear: 2022
+            cxx_std: 17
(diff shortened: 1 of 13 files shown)


src/rawtoaces_idt/rta.cpp
```diff
@@ -1028,13 +1028,15 @@ void Idt::loadCMF( const string &path )
...
+            uint16_t wl = atoi( ( row.first ).c_str() );
 
-            if ( _cmf[i]._wl < 380 || _cmf[i]._wl % 5 )
+            if ( wl < 380 || wl % 5 )
                 continue;
-            else if ( _cmf[i]._wl > 780 )
+            else if ( wl > 780 )
                 break;
 
+            _cmf[i]._wl = wl;
```
soswow, line 1038: I don't know what this is doing, but I can see how previously we would change _wl value before all the checks, where is now we apply it only if checks are passed. I assume that is intentional and fixes some problem?

.github/workflows/ci.yml
```diff
@@ -9,6 +9,75 @@ on:
     #branches: [ "master" ]
 
 jobs:
+  aswf_old:
```
soswow, line 12: Is that like a backup of previous aswf: which you've modified? How would this aswf_old be used?

src/rawtoaces_idt/rta.cpp
antond-weta: The original code first reads the wavelength into the next item of the array and then checks if it is in bounds. It can crash if the file contains more spectral samples than the tool expects (81). There are actually other places with the same bug, but this one does crash periodically as the cmd file does contain more samples.

PR #147, opened 2025-04-21
License update

aforsythe:
Update rawtoaces license to Apache 2.0


THIRD-PARTY.md
```diff
@@ -1,91 +1,13 @@
-#LICENSES
+See LICENSE for the main open source license of original code written for the rawtoaces project.
 
-1. [RAW TO ACES LICENSE](#raw-to-aces-license)
-2. [ACES CONTAINER LICENSE](#aces-container-license)
-3. [LIBRAW LICENSES](#libraw-licenses)
-4. [CERES SOLVER LICENSE](#ceres-solver-license)
-5. [ILMBASE and OPENEXR LICENSES](#ilmbase-and-openexr-licenses)
+The remainder of this file reproduces the open source licensing details of other projects that have been imported, incorporated into, or derived into parts of rawtoaces.
```
lgritz: This file only needs to cover projects whose code is reproduced IN THIS REPOSITORY. I believe that rawtoaces uses libraw, Imath, and OpenEXR only as link dependencies but does not copy any of its source code here, so they do not need to be mentioned here.

unittest/testSpst.cpp
```diff
@@ -1,56 +1,5 @@
...
-// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
-// THE POSSIBILITY OF SUCH DAMAGE.
-//
-// WITHOUT LIMITING THE GENERALITY OF THE FOREGOING, THE ACADEMY
-// SPECIFICALLY DISCLAIMS ANY REPRESENTATIONS OR WARRANTIES WHATSOEVER
-// RELATED TO PATENT OR OTHER INTELLECTUAL PROPERTY RIGHTS IN THE ACADEMY
-// COLOR ENCODING SYSTEM, OR APPLICATIONS THEREOF, HELD BY PARTIES OTHER
-// THAN A.M.P.A.S., WHETHER DISCLOSED OR UNDISCLOSED.
-///////////////////////////////////////////////////////////////////////////
+// SPDX-License-Identifier: apache-2.0
```
lgritz: I am honestly not sure if this is case sensitive ("Apache-2.0") or not. In order to have them be machine readable, that is. It might be best to use the standard capitalization, just to be safe.

lgritz (commented):
This looks great, exactly what we need.

See my comment about how the THIRD-PARTY file is actually including more than it needs to. In particular, I'm worried that this confusingly looks like you've incorporated libraw code into this project, and since it uses LGPL, that means that this whole project would need to be LGPL in order to be distributed. But I believe that's not actually what this project does -- if rawtoaces is merely using libraw as a dependency that is dynamically linked, it doesn't affect any of the other licensing, so we shouldn't take any chances that we give the impression that it does.

lgritz:
> Do we also need to put a license in the data files as well? Specifically, does the data in data/cmf/cmf_1931.json belong to CIE, and the data in data/illuminant/iso7589_stutung_380_780_5.json belong to ISO?

If the data file is a format that has provision for some kind of "comment" or other way to unobtrusively give an attribution, then yes, if possible, it should say where it comes from and who owns it. Just like what Alex did with the source code files here, the comment can just be 2 lines: one acknowledging the rights holder, and a second with the `SPDX-License-Identifier:` tag that identifies the license. This makes it easy for SBOM (software bill of materials) tools, including the one we use annually to scan the ASWF projects for license issues, to quickly inventory all the files and know what mix of licenses are represented.

Some data files don't have any way to be so marked, and in that case, it's ok to omit it but maybe mention it somewhere in the project documentation.

antond-weta:
Yes, there is the "license" field in the json files, which is currently set to null. The data in cmf_1931.json seems to be identical to https://files.cie.co.at/CIE_xyz_1931_2deg.csv, which is distributed under "CC BY-SA 4.0" according to the metadata file https://files.cie.co.at/CIE_xyz_1931_2deg.csv_metadata.json. This probably makes the json file a derivative work, it should also be under "CC BY-SA 4.0"? 

There are also some data blobs in the source files, which probably should be excluded from the Apache-2. For example, this https://github.com/AcademySoftwareFoundation/rawtoaces/blob/4ac227ded51ca51d746096041cc0bce14f41e584/include/rawtoaces/define.h#L186 is, I believe, the same data as https://cie.co.at/datatable/components-relative-spectral-distribution-daylight

lgritz:
> This probably makes the json file a derivative work, it should also be under "CC BY-SA 4.0"?

Sounds correct to me. Look up the spdx code for that license, I'm sure it has some standard way to say it, all the CC licenses are in there somewhere. https://spdx.org/licenses/


> There are also some data blobs in the source files, which probably should be excluded from the Apache-2. For example, this

Yes, I think it is sufficient to say in a comment, right at the start of the table, "The data in this table is taken from ..." and put the same 2 lines we put at the head of these files identifying the rights holder and license. I think that should make it fairly clear that it's just referring to the table data and not the rest of the file, while still making it so that the license scanners will correctly find it.

unittest/testSpst.cpp
aforsythe: updated in d5e80015c8bd7834fbffec5c739ec3c87251914d

THIRD-PARTY.md
aforsythe: Is the proposed resolution to simply remove THIRD-PARTY.md?

lgritz: I was only proposing that you remove the mention of packages that are merely build-time dependencies, and only use THIRD-PARTY to reproduce licenses for code "imported" into this repo from other projects and therefore not covered by our Apache-2.0. 

If there is no imported code in this repo at all, then there is no need for a THIRD-PARTY file.

aforsythe: @antond-weta can you confirm these are merely build-time dependencies?

antond-weta: Yes I can. There is no third-party code in the repository.

antond-weta:
> Yes, I think it is sufficient to say in a comment, right at the start of the table

Not a lawyer, but embedding of CC BY-SA 4.0 material may not be allowed, as per https://wiki.creativecommons.org/wiki/License_Versions#Application_of_effective_technological_measures_by_users_of_CC-licensed_works_prohibited. Distributing rawtoaces in binary form would restrict the users' access to the original data. 

It won't be hard to split out the tables into separate JSON files, if needed.

That is assuming that the data is indeed under CC BY-SA 4.0. I do think we should clarify the ownership/license of each piece of data.

lgritz:
@antond-weta We can check with LF legal if you want, but my guess is that while it would be nice to have a comment with a citation for the original source of the data (1931 blah blah), you probably don't need any license attribution there because the raw data in this small table is just a widely available list of facts, not "expressive content" that would be subject to copyright.

aforsythe:
> Do we also need to put a license in the data files as well? Specifically, does the data in data/cmf/cmf_1931.json belong to CIE, and the data in data/illuminant/iso7589_stutung_380_780_5.json belong to ISO?

We'd previously talked about moving the data here : https://github.com/AcademySoftwareFoundation/rawtoaces-data

aforsythe:
> There are also some data blobs in the source files

Seems like we should reference external files and make variable.  Hard coding this stuff into the source doesn't feel right.

antond-weta:
Should we go ahead with this PR, and I'll make the changes removing the data blobs and switching to using https://github.com/AcademySoftwareFoundation/rawtoaces-data as a separate PR?

antond-weta (approved):
Looks good

lgritz:
> Should we go ahead with this PR, and I'll make the changes removing the data blobs and switching to using https://github.com/AcademySoftwareFoundation/rawtoaces-data as a separate PR?

I think so, yes. We can continue to revise to address any concerns about the data blobs (which, in any case, are made no worse by anything in this PR, so there's no need to hold it up).

antond-weta:
> We'd previously talked about moving the data here : https://github.com/AcademySoftwareFoundation/rawtoaces-data

@aforsythe, rawtoaces-data is currently a private repo. Does it require more work on your side, or can we make it public?

PR #149, opened 2025-07-01
switch to rawtoaces_data repo

antond-weta:
Switch to using external repo for camera data.


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.11)
```
soswow, line 1: README still says 3.10 is a min. version

CMakeLists.txt
```diff
@@ -99,8 +99,37 @@ install( FILES "${PROJECT_BINARY_DIR}/RAWTOACESLibraryDepends.cmake" DESTINATION
...
 
+###############################################################################
+##
+##  Fetch data files from rawtoaces-data
+##
+include(FetchContent)
+
+FetchContent_Declare(
+  rawtoaces_data
+  GIT_REPOSITORY https://github.com/antond-weta/rawtoaces-data
```
soswow: `https://github.com/AcademySoftwareFoundation/rawtoaces-data`

soswow (approved):

README.md
```diff
@@ -245,225 +245,10 @@ By default, `rawtoaces` will determine the adopted white by finding the set of w
...
-    "required": [
-        "header",
-        "spectral_data"
-    ],
-    "type": "object"
-}
-```
+#### Spectral Datasets
+
+The spectral data have been moved to a separate repository [rawtoaces-data](https://github.com/AcademySoftwareFoundation/rawtoaces-data).
```
soswow: /nit I'd use more static language. One that describes the state, not how we got there. This one assumes reader knows that the dataset was imbedded into rawtoaces. So, the proposed format would be more like `The spectral data is located at a separate repository ... `

antond-weta:
converting this to a draft for now, as I'd like to make some more changes to the data repo

CMakeLists.txt
```diff
@@ -99,8 +99,34 @@ install( FILES "${PROJECT_BINARY_DIR}/RAWTOACESLibraryDepends.cmake" DESTINATION
...
+###############################################################################
+##
+##  Fetch data files from rawtoaces-data
+##
+include(FetchContent)
+
+FetchContent_Declare(
+  rawtoaces_data
+  GIT_REPOSITORY https://github.com/AcademySoftwareFoundation/rawtoaces-data
+  GIT_TAG 4db0acc872965ecdbbf578664be20abf17c18e8c # v0.1.0
```
soswow, line 104: looks like specific commit hash. before it was branch origin and name. Is there a way to specify `tag` itself somehow?

soswow: Seems like you can just put `v0.1.0`

lgritz: That's considered a supply chain security risk. Tags can be maliciously changed. Best practice is to use the full SHA hash. The comment says what the tag is, for easy human reference.

soswow: oh! that is cool. I didn't think about it.

README.md
soswow: Just want to make sure you are not planning on changing it. Since it was /nit I don't mind leaving it.

KelSolaar:
Sorry, coming late (I was on a break). Is there a specific reason for not using a Git sub-module here that is pulled when cloning the repo?

antond-weta:
> Sorry, coming late (I was on a break). Is there a specific reason for not using a Git sub-module here that is pulled when cloning the repo?

I don't have a specific reason, as this is the first time I'm doing this. I believe Kimball had advised me against using sub-modules here (unless I misunderstood badly).

antond-weta:
@KelSolaar I would suggest merging this as is for now, if you don't have any other concerns. I understand that we can switch to using git sub-modules later if needed, the change is self-contained and does not affect anything else.

A couple of arguments in support of the current solution off the top of my head:
- fetching the data-sources in cmake is more flexible, we can switch the repos conditionally; we may want, for example, pull additional camera models from the physlight repo if the matching cmake flag is set.
- fetching the data-sources in cmake we can check the compatible schema version and pick the corresponding tag

PR #151, opened 2025-07-17
fix cmake exports

antond-weta:
This fixes the export of cmake config files (RAWTOACESConfig.cmake, etc).
Also adds unit tests which mimic a client app linked against the installed libraries via the exported configs.

Most of the dependencies get resolved correctly, with these exceptions:
- I had to manually add Ceres on the client side, it seems the dependency gets resolved on the config side, but the Ceres::ceres alias does not get defined. 
- I also had to manually add libraw on the client side, on Windows only.


.github/workflows/ci.yml
```diff
@@ -145,17 +149,36 @@ jobs:
         # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
         # See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type
         run: >
-            cmake -B build -S .
-#            -DCMAKE_CXX_COMPILER=${{ matrix.cpp_compiler }}
-#            -DCMAKE_C_COMPILER=${{ matrix.c_compiler }}
-#            -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+          cmake -B build -S .
+          -D CMAKE_INSTALL_PREFIX=${{ steps.strings.outputs.build-output-dir }}/../install
```
soswow, line 153: where is this defined, btw?

.github/workflows/ci.yml
```diff
@@ -145,17 +149,36 @@ jobs:
...
```
soswow, line 154: It says you might need to add this to `aswf:` job
```suggestion
            - name: Set reusable strings
                # Turn repeated input strings (such as the build output directory) into step outputs. These step outputs can be used throughout the workflow file.
                id: strings
                shell: bash
                run: |
                  echo "build-output-dir=${{ github.workspace }}/build" >> "$GITHUB_OUTPUT"
```
on line 134

soswow: ignore me. `steps.strings.outputs.build-output-dir` has been already used in this step. I thought it's something new.

PR #152, opened 2025-07-18
Update GitHub org to AcademySoftwareFoundation

jfpanisset:
Although GitHub repo URLs will redirect from the previous organization a repo lived under,
update README.md to reflect the current org.


antond-weta (approved):
Looks good

PR #153, opened 2025-07-18
Replace hard coded value with CMAKE_LIB_DIR

jfpanisset:
This allows override of CMAKE_LIB_DIR to lib64 (for instance) on platforms where that's the preferred
destination for 64 bit DSOs.


jfpanisset:
CMAKE_LIB_DIR is inherited from top level CMakeLists.txt and defaults to "lib" as before.

PR #160, opened 2025-08-13
v2 - rename rawtoaces_idt to rawtoaces_core

antond-weta:


include/rawtoaces/rawtoaces_core.h
soswow, line 1: is that a breaking change? I don't fully understand how it all works, but it is possible someone has `rawtoaces/rta.h` import in their project?

Do we want make it breaking (because we will bump the version?) Or we want capability to stay for now?

In which case I understand this file has to stay as

```c++
// Compatibility header for legacy includes. Will be removed in a future release.

#pragma once

#if defined(_MSC_VER)
#pragma message("rawtoaces: <rawtoaces/rta.h> is deprecated; use <rawtoaces/rawtoaces_core.h>")
#else
#warning "rawtoaces: <rawtoaces/rta.h> is deprecated; use <rawtoaces/rawtoaces_core.h>"
#endif

#include <rawtoaces/rawtoaces_core.h>
```

include/rawtoaces/rawtoaces_core.h
```diff
@@ -7,7 +7,7 @@
 #include "define.h"
```
soswow: couple lines above should 
```c++
#ifndef _RTA_h__
#define _RTA_h__
```

be replaced with smth like 

```c++
#ifndef _RAWTOACES_CORE_H__
#define _RAWTOACES_CORE_H__
```

include/rawtoaces/rawtoaces_core.h
```diff
@@ -7,7 +7,7 @@
...
 
 #include <stdint.h>
-#include <libraw/libraw.h>
+#include <libraw.h>
```
soswow: why not both?

```c++
#if defined(__has_include)
#  if __has_include(<libraw/libraw.h>)
#    include <libraw/libraw.h>
#  elif __has_include(<libraw.h>)
#    include <libraw.h>
#  else
#    error "LibRaw headers not found. Ensure libraw is installed and include paths are set."
#  endif
#else
#  include <libraw/libraw.h>
#endif
```

you'd need a change in `src/rawtoaces_core/CMakeLists.txt` to go along the way.

> LibRaw is included in two common ways across systems:
#include <libraw/libraw.h> expects the compiler to search the parent include dir (e.g., /usr/local/include).
#include <libraw.h> expects the compiler to search the subdir itself (e.g., /usr/local/include/libraw).

> Without adding both paths here, one of the include styles would fail in environments where only the other path is provided.

src/rawtoaces_core/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+    target_link_libraries( ${RAWTOACES_CORE_LIB} PUBLIC Ceres::ceres )
+else ()
+    target_include_directories(${RAWTOACES_CORE_LIB} PUBLIC ${CERES_INCLUDE_DIRS})
+    target_link_libraries(${RAWTOACES_CORE_LIB} PUBLIC ${CERES_LIBRARIES})
+endif ()
+
+if ( LIBRAW_CONFIG_FOUND )
+    target_link_libraries ( ${RAWTOACES_CORE_LIB} PUBLIC libraw::raw )
+else ()
+    target_include_directories(${RAWTOACES_CORE_LIB} PUBLIC ${libraw_INCLUDE_DIR} )
```
soswow: this one change I've mentioned in prev. comment is on this line:

```c++
    get_filename_component(_libraw_parent_inc ${libraw_INCLUDE_DIR} DIRECTORY)
    target_include_directories(${RAWTOACES_CORE_LIB} PUBLIC ${_libraw_parent_inc} ${libraw_INCLUDE_DIR} )
```

I understand it would include prent of where libraw lives as well to search paths list

src/rawtoaces_core/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+    SOVERSION ${RAWTOACES_MAJOR_VERSION}.${RAWTOACES_MINOR_VERSION}.${RAWTOACES_PATCH_VERSION}
+    VERSION ${RAWTOACES_VERSION}
+)
+
+install( TARGETS ${RAWTOACES_CORE_LIB}
+    EXPORT RAWTOACESTargets
+    LIBRARY DESTINATION ${INSTALL_LIB_DIR}
+    INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+    PUBLIC_HEADER DESTINATION include/rawtoaces
+)
```
soswow: Do you want to add 

```c++
add_library(rawtoaces_idt INTERFACE)
target_link_libraries(rawtoaces_idt INTERFACE ${RAWTOACES_CORE_LIB})
install( TARGETS rawtoaces_idt EXPORT RAWTOACESTargets )
```

for backward compatibility so that those who previously linked `rawtoaces_idt` would not blow up?

Unless we want to make it a breaking change?

include/rawtoaces/rawtoaces_core.h
soswow: I forgot we are going to remove libraw in the next PR anyway. so, prob. ignore then

antond-weta:
> I don't see a changelog in this project, so maybe at least in README we can add something like
> 
> ```
> Note: The legacy header `rawtoaces/rta.h` and target name `rawtoaces_idt` are deprecated. 
> Use `rawtoaces/rawtoaces_core.h` and the `rawtoaces_core` target. 
> A compatibility header and exported interface target `rawtoaces_idt` are provided for a transitional period.
> ```
> 
> in case we keep things backward compatible. Or other text if we intend to break it.

the changelog is in docs/CHANGES.md. I have created it for v1.1. It doesn't contain any of the v2.0 notes yet.

soswow (approved):

PR #161, opened 2025-08-13
replace boost::filesystem with std::filesystem

antond-weta:


build_scripts/install_deps_windows.bash
```diff
@@ -8,6 +8,6 @@ vcpkg install \
     imath:x64-windows \
     boost-system:x64-windows \
     boost-foreach:x64-windows \
-    boost-filesystem:x64-windows \
+    boost-system:x64-windows \
```
soswow: duplicate of 9th line

CMakeLists.txt
soswow, line 1: Do we want to make it blow up if 17 requirement is not met:

On Line 3:
``` 
if(NOT DEFINED CMAKE_CXX_STANDARD)
  set(CMAKE_CXX_STANDARD 17)
endif()
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
```

configure.cmake
```diff
@@ -6,10 +6,9 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_INSTALL_PREFIX}/share/CMake"
 find_package ( AcesContainer CONFIG REQUIRED )
 find_package ( Eigen3        CONFIG REQUIRED )
 find_package ( Imath         CONFIG REQUIRED )
-find_package ( Boost                REQUIRED
+find_package ( Boost         CONFIG REQUIRED
     COMPONENTS
         system
-        filesystem
         unit_test_framework
 )
```
soswow, line 13: What do you of this:

The goal is to prefer CMake package config (vcpkg/Conan), fall back to FindBoost for distro installs

```
find_package ( Boost         CONFIG QUIET
    COMPONENTS
        system
        unit_test_framework
)
if (NOT Boost_FOUND)
    find_package ( Boost REQUIRED
        COMPONENTS
            system
            unit_test_framework
    )
endif()
```

antond-weta: I'm not too worried about this. Boost is going away in the next couple of PRs. Definitely before the next release.

CMakeLists.txt
antond-weta: Sure, why not. I'm using some of C++20 features in my v2 branch, so this is getting complicated.

build_scripts/install_deps_windows.bash
antond-weta: done

CMakeLists.txt
antond-weta: done

soswow (approved):

PR #162, opened 2025-08-14
v2 - remove the dependency on libraw from the core library

antond-weta:
This removes the libraw data structure being passed to the core library and replaces it with a dedicated structure rta::core::Metadata.

This changes the public interface, so is not backward compatible with v1.x, has to be a major update.

### DIFF
--- a/include/rawtoaces/acesrender.h
+++ b/include/rawtoaces/acesrender.h
@@ -7,6 +7,7 @@
 #include <rawtoaces/rta.h>
 
 #include <unordered_map>
+#include <libraw/libraw.h>
 
 using namespace rta;
 
@@ -72,7 +73,7 @@ private:
     const AcesRender &operator=( const AcesRender &acesrender );
 
     char                     *_pathToRaw;
-    Idt                      *_idt;
+    core::Idt                *_idt;
     libraw_processed_image_t *_image;
     LibRawAces               *_rawProcessor;
 
--- a/src/rawtoaces_idt/rta.cpp
+++ b/src/rawtoaces_idt/rta.cpp
@@ -1628,7 +1628,7 @@ DNGIdt::findXYZtoCameraMtx( const vector<double> &neutralRGB ) const
         lerror =
             mir - ccttoMired( XYZToColorTemperature( mulVector(
                       invertV( XYZtoCameraWeightedMatrix( mir, mir1, mir2 ) ),
-                      _metadata.neutralRGB ) ) );
+                      neutralRGB ) ) );
 
         if ( std::fabs( lerror - 0.0 ) <= 1e-09 )
         {
--- a/src/rawtoaces_idt/rta.cpp
+++ b/src/rawtoaces_idt/rta.cpp
@@ -13,6 +13,9 @@ using namespace ceres;
 
 namespace rta
 {
+namespace core
+{
+
 Illum::Illum()
 {
     _inc = 5;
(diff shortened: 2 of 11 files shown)


src/rawtoaces_idt/rta.cpp
```diff
@@ -1688,7 +1628,7 @@ DNGIdt::findXYZtoCameraMtx( const vector<double> &neutralRGB ) const
         lerror =
             mir - ccttoMired( XYZToColorTemperature( mulVector(
                       invertV( XYZtoCameraWeightedMatrix( mir, mir1, mir2 ) ),
-                      _neutralRGBDNG ) ) );
+                      _metadata.neutralRGB ) ) );
```
soswow: should this be coming from argument? so `neutralRGB`

unittest/testDNGIdt.cpp
```diff
@@ -39,15 +41,55 @@ BOOST_AUTO_TEST_CASE( TestIDT_RobertsonLength )
...
+    metadata.calibration[0].xyz2rgbMatrix.resize( 9 );
+    metadata.calibration[1].xyz2rgbMatrix.resize( 9 );
+    metadata.calibration[0].cameraCalibrationMatrix.resize( 9 );
+    metadata.calibration[1].cameraCalibrationMatrix.resize( 9 );
+
+    const libraw_colordata_t &color = libraw.imgdata.rawdata.color;
+
+#if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION( 0, 20, 0 )
+    metadata.baselineExposure =
+        static_cast<double>( color.dng_levels.baseline_exposure );
```
soswow, line 65: should it be `unsigned short` ?

unittest/testDNGIdt.cpp
```diff
@@ -56,7 +98,9 @@ BOOST_AUTO_TEST_CASE( TestIDT_XYZToColorTemperature )
     int ret = rawProcessor.open_file( ( pathToRaw.string() ).c_str() );
     ret     = rawProcessor.unpack();
```
soswow, line 99: do you want to put `BOOST_REQUIRE_EQUAL( ret, LIBRAW_SUCCESS );` after each line where `ret = ` is assigned? (so it would be open_file and upack. For early clear failure.

include/rawtoaces/acesrender.h
```diff
@@ -7,6 +7,7 @@
 #include <rawtoaces/rta.h>
 
 #include <unordered_map>
+#include <libraw/libraw.h>
 
 using namespace rta;
```
soswow, line 12: suggestion is to remove this namespace and replace those instances of `core::Idt` with `rta::core::Idt` here and in .cpp file
> to avoid polluting consumers’ namespaces from a public header. Public headers should not inject namespaces.

include/rawtoaces/acesrender.h
```diff
@@ -72,7 +73,7 @@ class AcesRender
     const AcesRender &operator=( const AcesRender &acesrender );
```
soswow, line 74: Suggestion is to move impl. bit of LibRawAces into .cpp
so lines 17-24 replace with `class LibRawAces;` and move all this into .cpp instead

src/rawtoaces_util/acesrender.cpp
soswow, line 1: is there a need for `#include <libraw/libraw.h>` in here as well? to 
> avoids those ODR/incomplete-type issues and matches how the types are actually defined by LibRaw.

include/rawtoaces/acesrender.h
antond-weta: Yes, but this is unrelated to the change. Trying to keep the scope of each PR limited. 
I generally dislike the `using namespace` thing, we should get rid of those in a separate PR, at least in the public headers.

soswow (approved):

PR #164, opened 2025-08-21
v2 - replace boost::json with nlohmann_json

antond-weta:
This replaces the boost::json with nlohmann_json, and adds a first-class citizen data class for the spectral data in the format we store the spectral curves in https://github.com/AcademySoftwareFoundation/rawtoaces-data

This is a breaking change, supposed to go live in the next major release.


.github/workflows/ci.yml
```diff
@@ -218,18 +222,22 @@ jobs:
...
+#          - os: windows-latest
+#            c_compiler: gcc
+#          - os: windows-latest
+#            c_compiler: clang
+#          - os: ubuntu-latest
+#            c_compiler: cl
+#          - os: macos-latest
+#            c_compiler: cl
+#          - os: macos-latest
+#            c_compiler: gcc
```
soswow: Are we planning on uncommenting this sometimes very soon? Can we delete the code instead of commenting out? (there is always git to bring back memories)

include/rawtoaces/define.h
```diff
@@ -97,8 +47,6 @@
 typedef float  float32_t;
 typedef double float64_t;
 
-using namespace std;
```
soswow, line 100: 👍

unittest/test_SpectralData.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+#include <filesystem>
+#include <boost/test/tools/floating_point_comparison.hpp>
+
+#include <rawtoaces/mathOps.h>
+#include <rawtoaces/rawtoaces_core.h>
+
+#define DATA_PATH "../_deps/rawtoaces_data-src/data/"
```
soswow, line 17: /nit being duplicated all over the place. Could be 
`add_definitions(-DDATA_PATH="${PROJECT_BINARY_DIR}/_deps/rawtoaces_data-src/data/")`

with this line here (And other usages) to be
```
#ifndef DATA_PATH
#    error "DATA_PATH not defined by CMake"
#endif
```

include/rawtoaces/spectral_data.h
```diff
@@ -0,0 +1,155 @@
...
+    /// - returns: the sum of all elements in `values`.
+    double integrate();
+
+    /// Find the maximum element in `values`
+    /// - returns: the maximum element in `values`.
+    double max() const;
+};
+
+/// A data-class for storing spectral data, based on the file format used in
+/// [rawtoaces-data](https://github.com/AcademySoftwareFoundation/rawtoaces-data).
```
soswow, line 97: Just curious what would happen if we decide to change this type. Would it be complex to do since it is a copy pasted from a source where data is coming from? (My mental picture is incomplete, I most likely saying silly things)

include/rawtoaces/util/data_adapters/spectral_database.h
```diff
@@ -0,0 +1,48 @@
+// Copyright Contributors to the rawtoaces project.
+// SPDX-License-Identifier: Apache-2.0
+// https://github.com/AcademySoftwareFoundation/rawtoaces
```
soswow: I noticed inconcistency with how these CC headers are among files. Some have if in different order, and others don't have a url. Not sure how important.

src/rawtoaces/main.cpp
```diff
@@ -88,13 +89,13 @@ int main( int argc, char *argv[] )
...
-        string output;
-        size_t pos = raw.rfind( '.' );
+        std::string output;
+        size_t      pos = raw.rfind( '.' );
         if ( pos != std::string::npos )
         {
             output = raw.substr( 0, pos );
         }
-        output += "_aces.exr";
+        output += "_aces2.exr";
```
soswow: sneaky. One of the breaking changes we would need to highlight in release docs.

include/rawtoaces/spectral_data.h
soswow: As in - are there ways to have single source of truth that is re-used in both places?

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -268,12 +105,10 @@ void Illum::calDayLightSPD( const int &cct )
 
     FORI( size )
     {
-        int index = s_series[0].wl + _inc * i;
+        int index = s_series[0].wl + inc * i;
         if ( index >= 380 && index <= 780 )
         {
-            _data.push_back( s01[i] + m1 * s11[i] + m2 * s21[i] );
-            if ( index == 550 )
```
soswow, line 275: what was this for before? We don't need this check any more? Looks like some hack for specific data set?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+        parse_string( d, bandwidth_corrected, "bandwidth_corrected" );
+
+        float prev_wavelength = -1;
+
+        std::vector<float>       wavelengths;
+        std::vector<std::string> keys;
+
+        nlohmann::json spectral_index = file_data["spectral_data"]["index"];
+
+        //        std::map<std::string, std::vector<std::string>> channels;
```
soswow: delete?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+        result += v;
+    return result;
+}
+
+double Spectrum::max() const
+{
+    double result = std::numeric_limits<float>::lowest();
+    for ( auto &v: values )
+        result = std::max( result, v );
+    return result;
```
soswow:
```cpp
if ( values.empty() )
    return 0;
return *std::max_element( values.begin(), values.end() );
```

can max_element be used here?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+}
+
+inline void
+parse_string( nlohmann::json &j, std::string &dst, const std::string &key )
+{
+    auto &v = j[key];
```
soswow, line 166: do we worry about key not being there at all?
Would it be `null`?
do we need to `!j.contains( key )` before this check?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+    if ( v.is_null() )
+        dst = "";
+    else
+        dst = v;
```
soswow, line 170: is there a need for `v.get<std::string>()` ?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+    transmission_geometry.erase();
+    bandwidth_FWHM.erase();
+    bandwidth_corrected.erase();
+    data.clear();
+
+    core::Spectrum::Shape shape;
+
+    try
+    {
+        std::ifstream  i( path );
```
soswow, line 199: is there a need for `if ( !i.is_open() )` check?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+        parse_string( h, creation_date, "document_creation_date" );
+        parse_string( h, comments, "comments" );
+        parse_string( h, license, "license" );
+
+        nlohmann::json &d = file_data["spectral_data"];
+        parse_string( d, units, "units" );
+        parse_string( d, reflection_geometry, "reflection_geometry" );
+        parse_string( d, transmission_geometry, "transmission_geometry" );
+        parse_string( d, bandwidth_FWHM, "bandwidth_FWHM" );
+        parse_string( d, bandwidth_corrected, "bandwidth_corrected" );
```
soswow, line 227: Is there a need for more verbose fail mode in case some of the fields are not present but considered required?

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+            {
+                float this_wavelength = std::stof( bin_wavelength );
+
+                if ( prev_wavelength != -1 )
+                {
+                    float new_step = this_wavelength - prev_wavelength;
+
+                    if ( shape.step != 0 && new_step != shape.step )
+                    {
+                        return false;
```
soswow, line 271: do we want to throw some warning like
```cpp
std::cerr << "Error: Inconsistent wavelength step detected. Expected: " 
                                  << shape.step << ", got: " << new_step << std::endl;
```

src/rawtoaces_core/spectral_data.cpp
```diff
@@ -0,0 +1,357 @@
...
+                    vv.second.reshape();
+                }
+            }
+        }
+    }
+    catch ( nlohmann::detail::parse_error error )
+    {
+
+        return false;
+    }
```
soswow, line 317: Same as above here. Maybe print out more details on why it failed
```cpp
    catch ( nlohmann::detail::parse_error &error )
    {
        std::cerr << "Error: JSON parsing failed: " << error.what() << std::endl;
        return false;
    }
    catch ( const std::exception &error )
    {
        std::cerr << "Error: Unexpected error during file loading: " << error.what() << std::endl;
        return false;
    }
```

soswow (approved):

unittest/test_SpectralData.cpp
antond-weta: Yeah, I would put it in a shared header, with the tabular data we use for testing. The unittest cleanup is coming as a separate PR soon

include/rawtoaces/spectral_data.h
antond-weta: I'm not clear what you mean. The data schema is defined in the rawtoaces-data. Do you mean the actual class and the data parser should live there as well?

I don't see this as a problem for now. Eventually we may need to modify the schema and introduce schema version checks in the parser, and maybe in other places in rawtoaces. But since there is only one schema version, and at is being used in all the files in the database, we are fine.

include/rawtoaces/util/data_adapters/spectral_database.h
antond-weta: Oh, this one is from another repo of mine, pushed by mistake. This file is not needed. Removed.

src/rawtoaces_core/rawtoaces_core.cpp
antond-weta: The 550nm thingy? That was stored by never used, probably a leftover code.

src/rawtoaces_core/spectral_data.cpp
antond-weta: deleted.

soswow (approved):

src/rawtoaces_core/spectral_data.cpp
antond-weta: nah, since de don't use the function with an arbitrary key (currently). added an assert

src/rawtoaces_core/spectral_data.cpp
antond-weta: This is tricky. For a camera the "manufacturer" and "model" are required. For an illuminant the "illuminant" is required. We parser doesn't know what the file is intended to be used for. The 3 things mentioned here are used for finding a camera or an illuminant. If a json has those missing, it is going to be skipped. Apart from these things rawtoaces does not require anything. Also the schema says that the "index" is required. But it actually doesn't mention that the number and the order of entries must match the "data".

src/rawtoaces_core/spectral_data.cpp
antond-weta: yep

src/rawtoaces_core/spectral_data.cpp
antond-weta: yep

src/rawtoaces_core/spectral_data.cpp
soswow: hm. your assertion `assert( !key.empty() );` checks for `key` (string) to be not empty. right?
Where I Was talking about making sure that this func ready for the case when object (`j`) doesn't have `key` at all.

.github/workflows/ci.yml
antond-weta: deleted

src/rawtoaces/main.cpp
antond-weta: used this for debugging, removed

src/rawtoaces_core/spectral_data.cpp
antond-weta: yes, done.

src/rawtoaces_core/spectral_data.cpp
soswow: I can see how what I said initially wasn't clear.

do we worry about key not being there **[in the object]** at all?
Would it [v] be null?

^ with missing bits that would make it clearer. sry

antond-weta: Oh, that is what this whole function is for. It works fine, as not all the files in the DB have all these properties set.

src/rawtoaces_core/spectral_data.cpp
antond-weta: It is being cast implicitly. Seems to be working fine.

soswow (approved):

PR #166, opened 2025-08-24
Changes to timer util private variable and unit tests

soswow:


src/rawtoaces_util/usage_timer.cpp
```diff
@@ -23,40 +23,34 @@ void UsageTimer::reset()
...
-        gettimeofday( &start_timeval, NULL );
-        _start_time = (double)start_timeval.tv_sec * 1000.0 +
-                      (double)start_timeval.tv_usec / 1000.0;
+        gettimeofday( &_start_timeval, NULL );
 #else
-        LARGE_INTEGER unit, start_timeval;
-        QueryPerformanceCounter( &start_timeval );
-        QueryPerformanceFrequency( &unit );
-        _start_time =
-            (double)start_timeval.QuadPart * 1000.0 / (double)unit.QuadPart;
```
soswow, line 35: instead of calculating ms twice it is done once in print() with raw value stored.

unittest/testUsageTimer.cpp
```diff
@@ -0,0 +1,231 @@
...
+    
+    // Extract timing values
+    float time1 = extractTimeFromOutput(output1);
+    float time2 = extractTimeFromOutput(output2);
+    
+    // Timer1 should show longer time (started earlier)
+    BOOST_CHECK_GT(time1, time2);
+    // Timer1 should be around 15ms, Timer2 around 5ms
+    BOOST_CHECK_GT(time1, 10.0f);
+    BOOST_CHECK_LT(time2, 10.0f);
```
antond-weta: nit: if you added both high and low checks to either time, you'd have 2 exact copies of the code which could be split into a function.

antond-weta (approved):

unittest/testUsageTimer.cpp
soswow: I'd leave it as is. A bit of duplication in unit tests is ok. Helps see what we are doing a bit more clearly. helps with debugging and defining the intent.

PR #167, opened 2025-08-27
v2 - remove libraw and aces_container

antond-weta:
Changes:
- replaces direct Libraw calls with OIIO for reading raw images
- replaces AcesContainer with OIIO for writing exr images
- replaces all pixel operations with OIIO IBAs

Since not all Libraw flags are exposed in the raw input plugin of OIIO, these command line parameters have been removed:
 -P - bad pixel mask
 -K - dark frame
 -j - fuji-rotate
 -m - median filter
 -f - four-colour RGB
 -T - print Libraw-supported cameras
 -F - use big file
 -E - mmap-ed IO
 -s - image index in the file

We may want to reintroduce them on the OIIO side later, if required.

This is a breaking change and not supposed to be compatible with v1.x.

### DIFF
--- a/configure.cmake
+++ b/configure.cmake
@@ -2,11 +2,10 @@
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_INSTALL_PREFIX}/share/CMake")
 
-
 find_package ( nlohmann_json CONFIG REQUIRED )
-find_package ( AcesContainer CONFIG REQUIRED )
+find_package ( OpenImageIO   CONFIG REQUIRED )
 find_package ( Eigen3        CONFIG REQUIRED )
-find_package ( Imath         CONFIG REQUIRED )
+#find_package ( Imath         CONFIG REQUIRED )
 find_package ( Boost         CONFIG REQUIRED
     COMPONENTS
         system
@@ -17,14 +16,4 @@ if (RTA_CENTOS7_CERES_HACK)
     find_package ( Ceres MODULE REQUIRED )
 else ()
     find_package ( Ceres CONFIG REQUIRED )
-endif ()
-
-find_package (libraw CONFIG QUIET )
-
-if (libraw_FOUND )
-    message("STATUS LibRaw config found")
-    set ( LIBRAW_CONFIG_FOUND TRUE )
-else ()
-    message("WARNING LibRaw config not found, trying to find a module.")
-    find_package(libraw MODULE REQUIRED)
 endif ()
--- a/src/rawtoaces_util/acesrender.cpp
+++ b/src/rawtoaces_util/acesrender.cpp
@@ -194,8 +194,7 @@ bool collect_image_files(
 
         for ( auto filename2: it )
         {
-            if ( !check_and_add_file( filename2, curr_batch ) )
-                continue;
+            check_and_add_file( filename2, curr_batch );
         }
     }
     else
@@ -284,19 +283,20 @@ bool fetch_camera_make_and_model(
     return true;
 }
 
-vector<string> findFiles( string filePath, vector<string> searchPaths )
+std::vector<std::string> find_files(
+    const std::string &file_path, const std::vector<std::string> &search_paths )
 {
-    vector<string> foundFiles;
+    std::vector<std::string> found_files;
 
-    for ( auto &searchPath: searchPaths )
+    for ( auto &search_path: search_paths )
     {
-        string path = searchPath + "/" + filePath;
+        std::string full_path = search_path + "/" + file_path;
 
-        if ( std::filesystem::exists( path ) )
-            foundFiles.push_back( path );
+        if ( std::filesystem::exists( full_path ) )
+            found_files.push_back( full_path );
     }
 
-    return foundFiles;
+    return found_files;
 }
 
 bool configure_solver(
@@ -325,14 +325,15 @@ bool configure_solver(
     }
 
     std::vector<std::string> found_training_data =
-        findFiles( "training/training_spectral.json", directories );
+        find_files( "training/training_spectral.json", directories );
     if ( found_training_data.size() )
     {
         // loading training data (190 patches)
         solver.loadTrainingData( found_training_data[0] );
     }
 
-    std::vector<std::string> found_cmf_files = findFiles( "cmf/cmf_1931.json", directories );
+    std::vector<std::string> found_cmf_files =
+        find_files( "cmf/cmf_1931.json", directories );
     if ( found_cmf_files.size() )
     {
         solver.loadCMF( found_cmf_files[0] );
@@ -422,6 +423,25 @@ bool solve_matrix_from_illuminant(
     return true;
 }
 
+/// Check if an attribute of a given name exists
+/// and has the type we are expecting.
+const OIIO::ParamValue *find_and_check_attribute(
+    const OIIO::ImageSpec &imageSpec,
+    const std::string     &name,
+    OIIO::TypeDesc         type )
+{
+    auto attr = imageSpec.find_attribute( name );
+    if ( attr )
+    {
+        auto attr_type = attr->type();
+        if ( attr_type == type )
+        {
+            return attr;
+        }
+    }
+    return nullptr;
+}
+
 bool prepare_transform_spectral(
     const OIIO::ImageSpec            &imageSpec,
     const ImageConverter::Settings   &settings,
@@ -449,9 +469,15 @@ bool prepare_transform_spectral(
         }
         else
         {
-            auto attr = imageSpec.find_attribute( "raw:pre_mul" );
-            for ( int i = 0; i < 4; i++ )
-                wb_multipliers[i] = attr->get_float_indexed( i );
+            auto attr = find_and_check_attribute(
+                imageSpec,
+                "raw:pre_mul",
+                OIIO::TypeDesc( OIIO::TypeDesc::FLOAT, 4 ) );
+            if ( attr )
+            {
+                for ( int i = 0; i < 4; i++ )
+                    wb_multipliers[i] = attr->get_float_indexed( i );
+            }
         }
 
         if ( wb_multipliers[3] != 0 )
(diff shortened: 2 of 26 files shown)


src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +849,649 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
-    {
-        FORIJ( 4, 4 ) XYZ_acesrgb[i][j] = XYZ_acesrgb_4[i][j];
-        aces = mulVectorArray( aces, total, 4, XYZ_acesrgb );
-    }
-    else
+bool ImageConverter::save_image(
+    const std::string &output_filename, const OIIO::ImageBuf &buf )
+{
+    const float chromaticities[] = { 0.7347, 0.2653, 0,       1,
+                                     0.0001, -0.077, 0.32168, 0.33767 };
```
soswow, line 1414: Maybe some comment explaining what magic numbers mean?

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +849,649 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
+        if ( spectral_white_balance )
+        {
+            float user_mul[4];
 
-float *AcesRender::renderACES()
-{
-#ifdef P
-#    undef P
-#endif
+            for ( int i = 0; i < _WB_multipliers.size(); i++ )
```
soswow:
```suggestion
            for ( size_t i = 0; i < _WB_multipliers.size(); i++ )
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +849,649 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
-                }
-            }
-            catch ( std::exception const &e )
-            {
-                std::cerr << e.what() << std::endl;
-            }
+            std::cerr << std::endl
+                      << "Error: No matching light source. "
+                      << "Please find available options by "
+                      << "\"rawtoaces --valid-illum\"." << std::endl;
```
soswow, line 982: is it ok we are not existing with after the error and continue out of if condition?

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +849,649 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
-            {
-                printf( "White Balance calculation method is 3 - " );
-                printf(
-                    "Using white balance factors calculated by "
-                    "averaging a grey box ...\n" );
-            }
-
+        case Settings::WBMethod::Illuminant: {
+            std::string lower_illuminant =
+                OIIO::Strutil::lower( settings.illuminant );
```
soswow: should there be some more action here?

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +849,649 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
-            OUT.output_color      = 0;
-            OUT.use_camera_matrix = 0;
+        case Settings::WBMethod::Custom:
+            options.attribute(
+                "raw:user_mul",
+                OIIO::TypeDesc( OIIO::TypeDesc::FLOAT, 4 ),
+                settings.customWB );
 
-            if ( _opts.verbosity > 1 )
+            //            if ( _read_raw )
```
soswow, line 1135:
```suggestion
```
delete comment (as well as not needed { })

src/rawtoaces_util/acesrender.cpp
```diff
@@ -106,275 +96,560 @@ void usage( const char *prog )
...
-//	outputs:
-//      static AcesRender &  : the referece to the only instance
-//      of "AcesRender" class
+vector<string> findFiles( string filePath, vector<string> searchPaths )
+{
+    vector<string> foundFiles;
+
+    for ( auto &i: searchPaths )
+    {
+        string path = i + "/" + filePath;
```
soswow:
```suggestion
    for ( auto &searchPath: searchPaths )
    {
        string path = searchPath + "/" + filePath;
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -106,275 +96,560 @@ void usage( const char *prog )
...
+        return false;
+    }
+
+    std::vector<std::string> found =
+        findFiles( "training/training_spectral.json", directories );
+    if ( found.size() )
+    {
+        // loading training data (190 patches)
+        solver.loadTrainingData( found[0] );
+    }
```
soswow:
```suggestion
    std::vector<std::string> found_training_data =
        findFiles( "training/training_spectral.json", directories );
    if ( found_training_data.size() )
    {
        // loading training data (190 patches)
        solver.loadTrainingData( found_training_data[0] );
    }
```
Also, do we care to let user know that there can be more then one and we pick the first one?

src/rawtoaces_util/acesrender.cpp
```diff
@@ -106,275 +96,560 @@ void usage( const char *prog )
...
 
-    mutex mtx;
-    mtx.lock();
-    static AcesRender acesrender;
-    mtx.unlock();
+    found = findFiles( "cmf/cmf_1931.json", directories );
+    if ( found.size() )
+    {
+        solver.loadCMF( found[0] );
+    }
```
soswow:
```suggestion
    std::vector<std::string> found_cmf_files = findFiles( "cmf/cmf_1931.json", directories );
    if ( found_cmf_files.size() )
    {
        solver.loadCMF( found_cmf_files[0] );
    }
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -106,275 +96,560 @@ void usage( const char *prog )
...
+    if ( illuminant.empty() )
+    {
+        auto illuminant_files = collect_data_files( directories, "illuminant" );
+        solver.loadIlluminant( illuminant_files );
+    }
+    else
+    {
+        auto illuminant_files = collect_data_files( directories, "illuminant" );
+        solver.loadIlluminant( illuminant_files, illuminant );
+    }
```
soswow:
```suggestion
    auto illuminant_files = collect_data_files( directories, "illuminant" );
    if ( illuminant.empty() )
    {
        solver.loadIlluminant( illuminant_files );
    }
    else
    {
        solver.loadIlluminant( illuminant_files, illuminant );
    }
```

src/rawtoaces_util/acesrender.cpp
antond-weta: No, that's fine. We may want to add an option to train on a custom set in the future.

src/rawtoaces_util/acesrender.cpp
antond-weta: done

src/rawtoaces_util/acesrender.cpp
antond-weta: done

src/rawtoaces_util/acesrender.cpp
antond-weta: no, removed.

src/rawtoaces_util/acesrender.cpp
```diff
@@ -106,275 +96,601 @@ void usage( const char *prog )
...
+    {
+        std::cerr << "Input transform matrix:" << std::endl;
+        for ( auto &i: IDT_matrix )
+        {
+            for ( auto &j: i )
+            {
+                std::cerr << j << " ";
+            }
+            std::cerr << std::endl;
+        }
```
soswow:
```suggestion
        for ( auto &row: IDT_matrix )
        {
            for ( auto &col: row )
            {
                std::cerr << col << " ";
            }
            std::cerr << std::endl;
        }
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -106,275 +96,601 @@ void usage( const char *prog )
...
+            return false;
+        }
+
+        if ( settings.verbosity > 0 )
+        {
+            std::cerr << "White balance coefficients:" << std::endl;
+            for ( auto &i: WB_multipliers )
+            {
+                std::cerr << i << " ";
+            }
```
soswow:
```suggestion
            for ( auto &wb_multiplier: WB_multipliers )
            {
                std::cerr << wb_multiplier << " ";
            }
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +883,653 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
+                    user_mul );
 
-            if ( _opts.verbosity > 1 )
-            {
-                printf( "White Balance calculation method is 2 - " );
-                printf(
-                    "Using white balance factors calculate by "
-                    "averaging the entire image ...\n" );
+                _WB_multipliers.resize( 4 );
+                for ( size_t i = 0; i < 4; i++ )
```
soswow, line 1136:
```suggestion
                for ( int i = 0; i < 4; i++ )
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +883,653 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
-                printf( "Using user-supplied white balance factors ...\n" );
-            }
+        case Settings::WBMethod::Custom:
+            options.attribute(
+                "raw:user_mul",
+                OIIO::TypeDesc( OIIO::TypeDesc::FLOAT, 4 ),
+                settings.customWB );
 
+            _WB_multipliers.resize( 4 );
+            for ( size_t i = 0; i < 4; i++ )
```
soswow, line 1172:
```suggestion
            for ( int i = 0; i < 4; i++ )
```

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +883,653 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
         }
         else
         {
-            cout << "rendering NonDNG" << endl;
-            return renderNonDNG();
+            if ( !prepare_transform_nonDNG(
+                     imageSpec, settings, _IDT_matrix, _CAT_matrix ) )
+            {
+                std::cerr << "ERROR: the colour space transform has not been "
+                          << "configured properly (Adobe mode)." << std::endl;
```
soswow, line 1273: is that a copy paste or should be here? Adobe is the next case, where same message is used.

antond-weta: that is correct. the only difference between the two is the value in "raw:use_camera_matrix" set above.

soswow (approved):

configure.cmake
```diff
@@ -2,11 +2,10 @@
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_INSTALL_PREFIX}/share/CMake")
 
-
 find_package ( nlohmann_json CONFIG REQUIRED )
-find_package ( AcesContainer CONFIG REQUIRED )
+find_package ( OpenImageIO   CONFIG REQUIRED )
 find_package ( Eigen3        CONFIG REQUIRED )
-find_package ( Imath         CONFIG REQUIRED )
+#find_package ( Imath         CONFIG REQUIRED )
```
soswow, line 8: we (I) can go through commented out things and remove them in the end, when all changes for v2 are made

src/rawtoaces_util/acesrender.cpp
```diff
@@ -548,1535 +879,653 @@ int AcesRender::configureSettings( int argc, char *argv[] )
...
-    if ( !pixels )
-    {
-        fprintf( stderr, "\nThe pixels cannot be found. \n" );
-        exit( 1 );
-    }
-    else
-    {
-        for ( uint32_t i = 0; i < total; i += 3 )
-        {
-            pixels[i]     = clip( _wbv[0] * pixels[i] / min_wb, target );
```
soswow, line 1534: here clip was killed

PR #169, opened 2025-08-28
public interface cleanup

antond-weta:
remove all unnecessary stuff from the public headers


soswow:
I appreciate all the full words in naming (calc -> calculate, illum -> Illuminant, etc) 👏 thank you.
We would need to consider writing "how to migrate" guide and change docs for changed public APIs. Myaybe future PR

soswow (approved):

PR #173, opened 2025-08-29
cleanup public interfaces

antond-weta:
Changes:
- `define.h` and `mathOps.h` are not public anymore
- renamed `acesrender.h/cpp` to `image_converter.h/cpp`
- renamed `unittest/` to `tests/`
- use a precalculated D65->ACES CAT matrix instead of re-calculating it every time


src/rawtoaces_util/image_converter.cpp
```diff
@@ -1279,6 +1278,19 @@ bool ImageConverter::configure(
     return configure( imageSpec, options );
 }
 
+// TODO:
```
soswow, line 1283: what's the plan for it?

soswow (approved):

src/rawtoaces_util/image_converter.cpp
antond-weta: No plan. If somebody requires any of those features, they should log an issue and we'll implement that properly in OIIO. Just wanted to keep the list there for reference.

PR #178, opened 2025-09-02
simplify core API

antond-weta:


include/rawtoaces/rawtoaces_core.h
```diff
@@ -28,95 +28,104 @@ static const std::vector<std::vector<double> > CAT_D65_to_ACES = {
...
+    SpectralData observer;
+
+    /// The training set spectral data. Can be either assigned directly, or loaded
+    /// in-place place via `solver.training_data.load()`.
+    SpectralData training_data;
+
+    /// The constructor. Takes the database search path as an optional
+    /// parameter.
+    /// @param search_path optional database search path.
+    SpectralSolver( const std::vector<std::string> &search_path = {} );
```
soswow: Can we make it `search_paths` since it's a list of things?

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -173,7 +173,8 @@ void generate_illuminant(
 
 // ------------------------------------------------------//
 
-SpectralSolver::SpectralSolver()
+SpectralSolver::SpectralSolver( const std::vector<std::string> &search_path )
+    : _search_path( search_path )
```
soswow: don't forget about this one to be `_search_paths` as well if you happy with above suggestion

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -217,227 +218,201 @@ void scaleLSC( const SpectralData &camera, SpectralData &illuminant )
...
+    std::vector<std::string> result;
+
+    for ( const auto &directory: _search_path )
+    {
+        if ( std::filesystem::is_directory( directory ) )
+        {
+            std::filesystem::path type_path( directory );
+            type_path.append( type );
+            if ( std::filesystem::exists( type_path ) )
+            {
```
soswow, line 234: do we care to warn about something in else case of this and previous if?

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -217,227 +218,201 @@ void scaleLSC( const SpectralData &camera, SpectralData &illuminant )
...
+                auto it = std::filesystem::directory_iterator( type_path );
+                for ( auto filename2: it )
+                {
+                    auto p = filename2.path();
+                    if ( filename2.path().extension() == ".json" )
+                    {
+                        result.push_back( filename2.path().string() );
+                    }
+                }
```
soswow:
```suggestion
                for ( auto filename: it )
                {
                    auto path = filename.path();
                    if ( path.extension() == ".json" )
                    {
                        result.push_back( path.string() );
                    }
                }
```
I can see you've cut/paste that from other place. But why not improve it just a touch while we are here.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -364,8 +204,48 @@ bool prepare_transform_spectral(
...
+                                      camera_model + "'";
+        print_error( data_type );
+        return false;
+    }
+
+    success = solver.load_spectral_data(
+        "training/training_spectral.json", solver.training_data );
+    if ( !success )
+    {
+        print_error( "training data" );
```
soswow, line 226: should there be more info here? (and next ones)

include/rawtoaces/rawtoaces_core.h
antond-weta: tricky, I'd rather call it search_directories then. usually $PATH is singular, but contains multiple locations

src/rawtoaces_core/rawtoaces_core.cpp
antond-weta: done

include/rawtoaces/rawtoaces_core.h
antond-weta: done

soswow (approved):

PR #183, opened 2025-09-03
More consistency into naming in image_converter

soswow:


src/rawtoaces_util/image_converter.cpp
```diff
@@ -191,17 +192,17 @@ void print_data_error( const std::string &data_type )
 }
 
 bool prepare_transform_spectral(
-    const OIIO::ImageSpec            &imageSpec,
+    const OIIO::ImageSpec            &image_spec,
     const ImageConverter::Settings   &settings,
-    std::vector<double>              &WB_multipliers,
+    std::vector<double>              &solved_WB_multipliers,
     std::vector<std::vector<double>> &IDT_matrix,
     std::vector<std::vector<double>> &CAT_matrix )
```
antond-weta: WB_multipliers is both input and output param, whereas IDT_matrix and CAT_matrix are pure outputs

antond-weta (commented):
disagree with `solved_WB_multipliers`, everything else is fine

src/rawtoaces_util/image_converter.cpp
soswow: addressed

antond-weta (approved):

PR #184, opened 2025-09-04
Some refactoring including CameraInfo struct

soswow:
I started going through the code again adding comments and thins and I noticed this oppartunity to simplify how we carry over camera make and model all over the place together. I've added a struct for that. What do you think?

### DIFF
--- a/src/rawtoaces_util/image_converter.cpp
+++ b/src/rawtoaces_util/image_converter.cpp
@@ -17,30 +17,43 @@ namespace rta
 namespace util
 {
 
+/**
+ * Checks if a file path is valid for processing and adds it to a batch list if appropriate.
+ * 
+ * This function validates that the given path points to a regular file or symbolic link,
+ * filters out unwanted files (system files like .DS_Store and certain image formats like EXR and JPG),
+ * and adds valid file paths to the provided batch vector for further processing.
+ * 
+ * @param path The filesystem path to check
+ * @param batch Reference to a vector of strings to add valid file paths to
+ * @return true if the file was processed (either added to batch or filtered out), 
+ *         false if the file should be ignored
+ */
 bool check_and_add_file(
     const std::filesystem::path &path, std::vector<std::string> &batch )
 {
-    if ( std::filesystem::is_regular_file( path ) ||
-         std::filesystem::is_symlink( path ) )
-    {
-        static const std::set<std::string> ignore_filenames = { ".DS_Store" };
-        std::string                        filename = path.filename().string();
-        if ( ignore_filenames.count( filename ) > 0 )
-            return false;
-
-        static const std::set<std::string> ignore_extensions = {
-            ".exr", ".EXR", ".jpg", ".JPG", ".jpeg", ".JPEG"
-        };
-        std::string extension = path.extension().string();
-        if ( ignore_extensions.count( extension ) > 0 )
-            return false;
-
-        batch.push_back( path.string() );
-    }
-    else
+    bool is_regular_file = std::filesystem::is_regular_file( path ) ||
+                           std::filesystem::is_symlink( path );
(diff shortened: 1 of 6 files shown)


src/rawtoaces_util/image_converter.cpp
```diff
@@ -17,30 +17,43 @@ namespace rta
...
-        std::string extension = path.extension().string();
-        if ( ignore_extensions.count( extension ) > 0 )
-            return false;
-
-        batch.push_back( path.string() );
-    }
-    else
+    bool is_regular_file = std::filesystem::is_regular_file( path ) ||
+                           std::filesystem::is_symlink( path );
+    if ( !is_regular_file )
```
soswow, line 37: moved it here for early exit. less if inside if situation

src/rawtoaces_util/image_converter.cpp
```diff
@@ -17,30 +17,43 @@ namespace rta
...
+
+    static const std::set<std::string> ignore_filenames = { ".DS_Store" };
+    std::string                        filename = path.filename().string();
+    if ( ignore_filenames.count( filename ) > 0 )
+        return false;
+
+    static const std::set<std::string> ignore_extensions = { ".exr",
+                                                             ".jpg",
+                                                             ".jpeg" };
+    std::string extension = OIIO::Strutil::lower( path.extension().string() );
```
soswow, line 51: Using `OIIO::Strutil::lower` to remove extra extenstions from the list above.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -113,31 +130,25 @@ std::vector<std::string> database_paths()
...
-            end = path.size();
-
-        std::string pathItem = path.substr( pos, end - pos );
-
-        if ( find( result.begin(), result.end(), pathItem ) == result.end() )
-            result.push_back( pathItem );
-
-        pos = end + 1;
-    }
+    OIIO::Strutil::split( path, result, separator );
```
soswow, line 133: Using existing `OIIO::Strutil::split` thought as I understand we are loosing logic that was removing duplicates. Don't know if that was intentional and important part of this code. If so, this won't work.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -113,31 +130,25 @@ std::vector<std::string> database_paths()
...
+/// Get camera info (with make and model) from image metadata or custom settings.
+///
+/// Returns camera information using custom settings if provided, otherwise
+/// extracts from image metadata. Returns empty CameraInfo if required metadata is missing.
+///
+/// @param spec Image specification containing metadata
+/// @param settings Converter settings with optional custom camera info
+/// @return CameraInfo struct with make and model, or empty if unavailable
+rta::core::CameraInfo get_camera_info(
+    const OIIO::ImageSpec &spec, const ImageConverter::Settings &settings )
```
soswow, line 147: a bit more functional - you give sources of where to get camera info from, you get camera info out.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -359,10 +370,10 @@ bool prepare_transform_DNG(
...
     std::vector<std::vector<double>> &CAT_matrix )
 {
-    std::string camera_make  = settings.custom_camera_make;
-    std::string camera_model = settings.custom_camera_model;
-    if ( !fetch_camera_make_and_model( image_spec, camera_make, camera_model ) )
-        return false;
+    // TO Does this function even need the camera info?
+    // core::CameraInfo camera_info = get_camera_info( image_spec, settings );
+    // if ( camera_info.is_empty() )
+    //     return false;
```
soswow, line 375: was it used just for validation `settings` that it has camera settings? because camera_make wasn't used anywhere

PR #185, opened 2025-09-05
More refactoring and descriptions

soswow:


src/rawtoaces_util/image_converter.cpp
```diff
@@ -262,22 +276,25 @@ bool prepare_transform_spectral(
...
         else
         {
-            auto attr = find_and_check_attribute(
-                image_spec,
+            // Extract white balance from RAW metadata
+            auto attr = image_spec.find_attribute(
                 "raw:pre_mul",
-                OIIO::TypeDesc( OIIO::TypeDesc::FLOAT, 4 ) );
+                OIIO::TypeDesc(
+                    OIIO::TypeDesc::FLOAT, OIIO::TypeDesc::AGGREGATE::VEC4 ) );
```
soswow: Notice use of enums for some of them (OIIO::TypeDesc::AGGREGATE::VEC4)

src/rawtoaces_util/image_converter.cpp
```diff
@@ -347,38 +366,51 @@ bool prepare_transform_spectral(
...
     const OIIO::ImageSpec            &image_spec,
     const ImageConverter::Settings   &settings,
     std::vector<std::vector<double>> &IDT_matrix,
     std::vector<std::vector<double>> &CAT_matrix )
 {
-    std::string camera_make  = settings.custom_camera_make;
-    std::string camera_model = settings.custom_camera_model;
-    if ( !fetch_camera_make_and_model( image_spec, camera_make, camera_model ) )
-        return false;
-
```
soswow, line 366: I couldn't find a reason for this to exist. None of the values are used down below. Let me know if I should revert that.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -438,84 +476,99 @@ bool prepare_transform_DNG(
...
+    // Step 5: Clear CAT matrix (not used for DNG)
     // Do not apply CAT for DNG
     CAT_matrix.resize( 0 );
     return true;
 }
 
-bool prepare_transform_nonDNG(
-    const OIIO::ImageSpec            &image_spec,
-    const ImageConverter::Settings   &settings,
+void prepare_transform_nonDNG(
```
soswow, line 499: if you feel strongly about keeping it `bool` for consistency, let me know. I don't see much point

src/rawtoaces_util/image_converter.cpp
```diff
@@ -576,6 +629,7 @@ bool check_param(
         }
         else
         {
+            on_success(); // TODO Verify with Anton.
```
soswow: TODO Verify that lack of on_success in this return true case was a bug or intentional.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -794,6 +848,18 @@ void ImageConverter::init_parser( OIIO::ArgParse &arg_parser )
...
+std::vector<std::string> get_map_keys( const MapType &map )
+{
+    std::vector<std::string> keys;
+    for ( const auto &pair: map )
+    {
+        keys.push_back( pair.first );
+    }
+    return keys;
+}
+
```
soswow: I might be missing something. Not sure if there is an easier way of extracting keys from a Map.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -826,80 +891,70 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
     {
         std::cerr << std::endl
-                  << "Unsupported white balancing method: \"" << WB_method
-                  << "\"." << std::endl;
-
+                  << "ERROR: unsupported white balancing method '" << WB_method
+                  << "'. The following methods are supported: '"
+                  << OIIO::Strutil::join( get_map_keys( wb_methods ), "', '" )
+                  << "'." << std::endl;
         return false;
```
soswow: Two improvements in one. No need for repeating code in if/else + listing available options in error.
Integrating it with actuall help message might be even better.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -826,80 +891,70 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
-                      << "\"illuminant\", but no \"--illuminant\" parameter "
-                      << "provided. D55 will be used as default." << std::endl;
-            settings.illuminant = "D55"; // Set default illuminant
-        }
+        std::cerr << "Warning: the white balancing method was set to "
+                  << "\"illuminant\", but no \"--illuminant\" parameter "
+                  << "provided. D55 will be used as default." << std::endl;
+
+        std::string default_illuminant = "D55";
+        settings.illuminant            = default_illuminant;
```
soswow, line 959: just replacing a comment with a named variable.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1180,21 +1232,24 @@ bool ImageConverter::configure(
     options["raw:Demosaic"]           = settings.demosaic_algorithm;
     options["raw:threshold"]          = settings.denoise_threshold;
 
+    const OIIO::TypeDesc float_vec4_type_desc(
+        OIIO::TypeDesc::FLOAT, OIIO::TypeDesc::AGGREGATE::VEC4 );
+    const OIIO::TypeDesc int_vec4_type_desc(
+        OIIO::TypeDesc::INT, OIIO::TypeDesc::AGGREGATE::VEC4 );
+    const OIIO::TypeDesc float_vec2_type_desc(
+        OIIO::TypeDesc::FLOAT, OIIO::TypeDesc::AGGREGATE::VEC2 );
```
soswow: TODO Eithe rmove it up top outside all methods and use them in other methods or use inline in all the places.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1362,24 +1406,12 @@ bool ImageConverter::configure(
         }
         else
         {
-            if ( !prepare_transform_nonDNG(
-                     image_spec, settings, _IDT_matrix, _CAT_matrix ) )
-            {
-                std::cerr << "ERROR: the colour space transform has not been "
```
soswow, line 1368: this will never run

include/rawtoaces/image_converter.h
```diff
@@ -22,6 +22,19 @@ namespace util
 bool collect_image_files(
     const std::string &path, std::vector<std::vector<std::string>> &batches );
 
+struct CameraIdentifier
```
antond-weta: Nice! Can this live in the .cpp?

include/rawtoaces/image_converter.h
```diff
@@ -22,6 +22,19 @@ namespace util
...
+{
+    std::string make;
+    std::string model;
+
+    CameraIdentifier() = default;
+    CameraIdentifier( const std::string &_make, const std::string &_model )
```
antond-weta: this constructor is not strictly needed, you can initialise the struct members directly
camera_identifier = { "ARRI", "too_expensive_for_me" };

src/rawtoaces_util/image_converter.cpp
```diff
@@ -406,9 +440,12 @@ bool prepare_transform_DNG(
...
         }
 
+        // Extract camera calibration matrix
         auto key2         = "raw:dng:camera_calibration" + index_string;
-        auto matrix2_attr = find_and_check_attribute(
-            image_spec, key2, OIIO::TypeDesc( OIIO::TypeDesc::FLOAT, 16 ) );
+        auto matrix2_attr = image_spec.find_attribute(
+            key2,
+            OIIO::TypeDesc(
+                OIIO::TypeDesc::FLOAT, OIIO::TypeDesc::AGGREGATE::MATRIX44 ) );
```
antond-weta: have you confirmed that this works? that  float[4][4] has the correct order of items when read from float[16]? A zero or identity matrix is not a sufficient test

src/rawtoaces_util/image_converter.cpp
antond-weta: Honestly I can't remember the logic there. Should have put a comment. Could you try reproducing visiting that branch? Also, probably checking the size in the branch above where `return true;` is not necessary anymore, as ArgParse should enforce that.

src/rawtoaces_util/image_converter.cpp
antond-weta: good enough. there is a simpler way in c++20, but we're not there yet

alternatively, you could use 2 separate vectors for your keys and values. 

keep in mind that the map stores the data sorted. not a big problem, but you may get the "custom" listed first, not last.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -826,80 +891,70 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
-    else if ( WB_method == "box" )
-    {
-        settings.WB_method = Settings::WBMethod::Box;
-    }
-    else if ( WB_method == "custom" )
-    {
-        settings.WB_method = Settings::WBMethod::Custom;
-    }
-    else
+    static std::map<std::string, Settings::WBMethod> wb_methods = {
```
antond-weta: you may want it to const as well

src/rawtoaces_util/image_converter.cpp
antond-weta: or just don't. I'd rather see the type on the line I'm working on, rather than have to search for it. this saves 5 lines of code by making it harder to read.

src/rawtoaces_util/image_converter.cpp
soswow: killed it. 4 is cleaner

src/rawtoaces_util/image_converter.cpp
soswow: I Am killing this anyway, but I thought this is just a fency variable that holds value 16
```
enum AGGREGATE {
        SCALAR   =  1,  ///< A single scalar value (such as a raw `int` or
                        ///<   `float` in C).  This is the default.
        VEC2     =  2,  ///< 2 values representing a 2D vector.
        VEC3     =  3,  ///< 3 values representing a 3D vector.
        VEC4     =  4,  ///< 4 values representing a 4D vector.
        MATRIX33 =  9,  ///< 9 values representing a 3x3 matrix.
        MATRIX44 = 16   ///< 16 values representing a 4x4 matrix.
    };
```

src/rawtoaces_util/image_converter.cpp
soswow: killing it anyway.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -958,7 +1018,7 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
                     settings.custom_matrix[i][j] = i == j ? 1.0 : 0.0;
         } );
 
-    auto crop_box = arg_parser["crop_box-box"].as_vec<int>();
```
soswow, line 961: I assume thig was a bug? I've picke dit up accidentally

antond-weta: yes

src/rawtoaces_util/image_converter.cpp
```diff
@@ -191,35 +193,53 @@ void print_data_error( const std::string &data_type )
...
-    success = solver.find_camera( camera_make.c_str(), camera_model.c_str() );
+    success =
+        solver.find_camera( camera_identifier.make, camera_identifier.model );
     if ( !success )
     {
         const std::string data_type = "spectral data for camera make = '" +
-                                      camera_make + "', model = '" +
-                                      camera_model + "'";
+                                      camera_identifier.make + "', model = '" +
+                                      camera_identifier.model + "'";
```
antond-weta: I suspect we'll end up having to print out `make='Sony', model='who_cares'` in more than just this place, especially in higher verbosity modes. You may want to move this to `CameraIdentifier.description()`, or `CameraIdentifier.to_string()`

src/rawtoaces_util/image_converter.cpp
```diff
@@ -996,26 +1057,26 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
             settings.chromatic_aberration[i] = chromatic_aberration[i];
     }
 
-    auto demosaic_algorithm = arg_parser["demosaic"].get();
-    static std::set<std::string> demosaic_algorithms = {
+    auto                         demosaic       = arg_parser["demosaic"].get();
+    static std::set<std::string> demosaic_algos = {
```
antond-weta: nah, put that back

src/rawtoaces_util/image_converter.cpp
```diff
@@ -996,26 +1057,26 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
-    if ( demosaic_algorithms.count( demosaic_algorithm ) != 1 )
+    if ( demosaic_algos.count( demosaic ) != 1 )
     {
         std::cerr << std::endl
-                  << "ERROR: unsupported demosaicing algorithm '"
-                  << demosaic_algorithm
-                  << ". The following methods are supported: "
+                  << "ERROR: unsupported demosaicing algorithm '" << demosaic
+                  << "'. "
+                  << "The following algorithms are supported: "
```
antond-weta, line 1076: here is a good spot to use join()?

src/rawtoaces_util/image_converter.cpp
```diff
@@ -633,12 +633,40 @@ bool check_param(
...
         }
     }
 }
 
+template <typename T> bool contains( const std::vector<T> &vec, const T &value )
+{
+    return std::find( vec.begin(), vec.end(), value ) != vec.end();
+}
+
+static std::vector<std::string> supported_wb_methods = {
```
antond-weta: You could have put these static methods inside the `Settings`, close to there the enums are defined.

Or try something like this https://stackoverflow.com/a/31836401

src/rawtoaces_util/image_converter.cpp
soswow: yeah. that was accedental.

src/rawtoaces_util/image_converter.cpp
soswow: done.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -801,27 +869,21 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
-        std::cerr << std::endl
-                  << "The following illuminants are supported:" << std::endl;
-        for ( const auto &illuminant: illuminants )
-        {
-            std::cerr << std::endl << illuminant;
-        }
-        std::cerr << std::endl;
+        std::cout << std::endl
+                  << "The following illuminants are supported:" << std::endl
+                  << OIIO::Strutil::join( illuminants, "\n" ) << std::endl;
```
soswow, line 885: found this other place wher `join` can be used

src/rawtoaces_util/image_converter.cpp
```diff
@@ -801,27 +869,21 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
+        exit( 0 );
```
soswow, line 886: fixes the problem of showing help message for these two commands. Returning `false` would cause for programm to exit with `1` in main, which semantically incorrect.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -191,35 +204,52 @@ void print_data_error( const std::string &data_type )
...
-    success = solver.find_camera( camera_make.c_str(), camera_model.c_str() );
+    success =
+        solver.find_camera( camera_identifier.make, camera_identifier.model );
     if ( !success )
     {
-        const std::string data_type = "spectral data for camera make = '" +
-                                      camera_make + "', model = '" +
-                                      camera_model + "'";
+        const std::string data_type =
+            "spectral data for camera " + camera_identifier;
```
soswow, line 247: As suggested, I've implemented toString operator in CameraIdentifier

src/rawtoaces_util/image_converter.cpp
```diff
@@ -17,30 +17,64 @@ namespace rta
...
+    {
+        return "make: '" + make + "', model: '" + model + "'";
     }
-    else
+
+    friend std::string
+    operator+( const std::string &lhs, const CameraIdentifier &rhs )
+    {
+        return lhs + static_cast<std::string>( rhs );
+    }
```
soswow, line 38: Can't claim I know how it's done in c++. That's what I came up with so I can do `"mystring" + this_object`. I believe it would need smth like
```
// Stream insertion operator
    friend std::ostream& operator<<(std::ostream& os, const CameraIdentifier& camera) {
        return os << camera.make << " " << camera.model;
    }
```
if we want to use it in `"some string" << this_object` scenario. But I am not adding it now

antond-weta (approved):
looks good

PR #186, opened 2025-09-06
Fix wrong success execution

soswow:


src/rawtoaces_util/image_converter.cpp
```diff
@@ -643,7 +643,8 @@ bool check_param(
         }
         else
         {
-            on_success();
+            // Correct mode and no parameter provided.
```
antond-weta: Incorrect mode and no parameter. We don't want any mode-specific parameters parsed when we are in wrong mode.

PR #189, opened 2025-09-08
Add data-dir argument and auto matrix method

### LINKED ISSUE #180: add a command line parameter for the database search path
We should add a parameter (--data_path?) with the same functionality as the RAWTOACES_DATA_PATH environment variable. 
In case if both the command line parameter and environment variable are present, the command line should take priority.

### LINKED ISSUE #181: Add "auto" matrix method and make it default.
Currently the "spectral" method is default, which is the most accurate, but makes rawtoaces fail if it doesn't have the camera measurements in the database. This means that rawtoaces would fail for 99% images when called without any extra command line parameters. This is not very user friendly. 

Adding an "auto" method, which would use "spectral" if there are spectral measurements, falling back to "metadata" otherwise, should solve this problem.

mikaelsundell:
- Added --data-dir argument (Fixes #180)
- Added auto matrix method (Fixes #181)
- Add size checks for lambda array assignments


src/rawtoaces_util/image_converter.cpp
```diff
@@ -968,60 +993,69 @@ bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
...
-        settings.WB_method == Settings::WBMethod::Box,
-        [&]() {
-            for ( int i = 0; i < 4; i++ )
-                settings.WB_box[i] = WB_box[i];
-        },
-        [&]() {
-            for ( int i = 0; i < 4; i++ )
-                settings.WB_box[i] = 0;
-        } );
+    if ( WB_box.size() == 4 )
```
antond-weta: This check is not needed as it is performed inside of `check_param`, I assume you were fixing the crash we introduced recently, we do have a PR with a fix [#186](https://github.com/AcademySoftwareFoundation/rawtoaces/pull/186)

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1335,6 +1369,31 @@ bool ImageConverter::configure(
             return false;
     }
 
+    if ( settings.matrix_method == Settings::MatrixMethod::Auto )
```
antond-weta, line 1384: I suspect this will only work for the first input? You need to retain the original choice somewhere and do this check for each input.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -864,7 +877,15 @@ void ImageConverter::init_parser( OIIO::ArgParse &arg_parser )
 
 bool ImageConverter::parse_parameters( const OIIO::ArgParse &arg_parser )
 {
-    settings.database_directories = database_paths();
+    std::string data_dir = arg_parser["data-dir"].get();
+    if ( data_dir.size() )
+    {
+        OIIO::Strutil::split( data_dir, settings.database_directories, ":" );
```
antond-weta: Will this work on windows? See the logic in `database_paths()`. Perhaps it would be better to pass this string as a parameter to `database_paths()` so all pass resolving logic is in one place.

src/rawtoaces_util/image_converter.cpp
mikaelsundell: Yes, I was too fast — now removed.

src/rawtoaces_util/image_converter.cpp
mikaelsundell: Fixed in latest commit.

mikaelsundell:
The new changes are ready for review. Once they are approved, I’ll add tests to improve coverage and address the Codecov report.

src/rawtoaces_util/image_converter.cpp
mikaelsundell: Fixed in latest commit.

include/rawtoaces/image_converter.h
```diff
@@ -252,6 +259,17 @@ class ImageConverter
...
 
+    /// Process a batch of input files or directories.
+    /// This will iterate over each batch and call process_image() on every file.
+    /// @param batches
+    ///     Vector of batches, each batch is a vector of filenames.
+    /// @param empty
+    ///     Output flag that becomes true if no files were processed.
+    /// @result
+    ///     true if all files were processed successfully, false otherwise.
+    bool process_batch(
```
antond-weta: I like all other changes, but I find this one questionable. 
- Is this important enough to warrant an API change? It is just a loop, why can't it live on the client side?
- the name is misleading, it should be called `process_batches`
- the output `empty` flag looks kludgy for a public API

I prefer this logic being on the client side, because the user can decide how to treat failures - is it important that all files succeed? shall we continue if a file failed? does a failed file means the whole sequence is broken, or a single batch? All these decisions should be on the client side.

mikaelsundell: Thanks, this is a very quick draft/ idea for discussion, in short, I moved it into the image converter to make it a part of the library (using settings.verbosity) etc. The empty flag is redundant as-is.  Guess we would need to provide some sort of status per file after a run to make it work, let's drop the idea for now and let the client do all batch processing?

antond-weta: Yes, let's drop it for now, we can return to this at a later stage. I'm anticipating the image processing pipeline being much longer in the future, with many more steps added. We may need to reconsider the error reporting and failure detection logic then. Please finish up the rest of the PR and we'll get it merged in.

soswow:
> The new changes, along with a draft for verbose output, are ready for review. Once they are approved, I’ll add tests to improve coverage and address the Codecov report.

See my other PR curretly open. I was adding those tests before I saw your PR. Just be mindfull of it.

include/rawtoaces/image_converter.h
mikaelsundell: Done, you could argue for having a bit more output even when not calling -v, like at least giving info about the current file being processed. Right now it's silent except for OIIO output for debug builds.

src/rawtoaces/main.cpp
```diff
@@ -40,17 +40,42 @@ int main( int argc, const char *argv[] )
...
+    // Process raw files
     bool empty  = true;
     bool result = true;
+
+    int file_index  = 0;
+    int total_files = 0;
+
+    for ( auto const &batch: batches )
+        total_files += static_cast<int>( batch.size() );
+
```
antond-weta, line 52: why not use `size_t` and avoid casting? this use case is exactly what `size_t` is for

src/rawtoaces_util/image_converter.cpp
```diff
@@ -378,9 +390,10 @@ bool prepare_transform_spectral(
 
     if ( settings.verbosity > 0 )
     {
-        std::cerr << "Input transform matrix:" << std::endl;
+        std::cerr << "Input Device Transform (IDT) matrix:" << std::endl;
```
antond-weta, line 393: The ACES group has stopped using the term IDT some time ago. I've been trying to minimise its usage here.

soswow:
`database_paths` is now fully unit tests in `tests/test_image_converter.cpp`. Now you can add/modify tests to cover new logic.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -869,7 +895,15 @@ void ImageConverter::init_parser( OIIO::ArgParse &arg_parser )
...
-    settings.database_directories = database_paths();
+    std::string data_dir = arg_parser["data-dir"].get();
+    if ( data_dir.size() )
+    {
+        settings.database_directories = database_paths( data_dir );
+    }
+    else
+    {
+        settings.database_directories = database_paths();
+    }
```
soswow, line 907: there are unit tests now for `list-cameras` and `list-illuminants`. They are the only but maybe the easest way you can test functionality above. Tests currently setup test-dit location via env. You could have one more test for cameras with same exact logic, but setup path with --data-dir and it would test new branch above.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1433,11 +1496,109 @@ bool ImageConverter::configure(
...
+                  << ( settings.output_dir.empty() ? "<same as input>"
+                                                   : settings.output_dir )
+                  << std::endl;
+        std::cerr << "  Overwrite: " << ( settings.overwrite ? "yes" : "no" )
+                  << std::endl;
+        std::cerr << "  Create dirs: "
+                  << ( settings.create_dirs ? "yes" : "no" ) << std::endl;
+        std::cerr << "  Verbosity: " << settings.verbosity << std::endl;
+    }
+
```
soswow, line 1603: I feel line all this above can live in own private function. Takes settings in and prints it out. `print_out_settings()` or something.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1433,11 +1496,109 @@ bool ImageConverter::configure(
...
+            case Settings::CropMode::Soft: std::cerr << "soft"; break;
+            case Settings::CropMode::Hard: std::cerr << "hard"; break;
+        }
+        std::cerr << std::endl;
+
+        if ( settings.crop_box[2] > 0 && settings.crop_box[3] > 0 )
+        {
+            std::cerr << "  Crop box: [" << settings.crop_box[0] << ", "
+                      << settings.crop_box[1] << ", " << settings.crop_box[2]
+                      << ", " << settings.crop_box[3] << "]" << std::endl;
```
soswow, line 1587: check out `OIIO::Strutil::join()`

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1433,11 +1496,109 @@ bool ImageConverter::configure(
...
+        {
+            std::cerr << "  Custom WB: [" << settings.custom_WB[0] << ", "
+                      << settings.custom_WB[1] << ", " << settings.custom_WB[2]
+                      << ", " << settings.custom_WB[3] << "]" << std::endl;
+        }
+
+        if ( settings.matrix_method == Settings::MatrixMethod::Custom )
+        {
+            std::cerr << "  Custom matrix:" << std::endl;
+            for ( int i = 0; i < 3; i++ )
```
soswow, line 1566: /nit  there is `FORI(3)` template available if you want

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1377,8 +1441,7 @@ bool ImageConverter::configure(
 
     bool spectral_white_balance =
         settings.WB_method == Settings::WBMethod::Illuminant;
-    bool spectral_matrix =
-        settings.matrix_method == Settings::MatrixMethod::Spectral;
+    bool spectral_matrix = matrix_method == Settings::MatrixMethod::Spectral;
```
soswow, line 1446: /nit `is_spectral_matrix`? otherwise one might think it's spectral matrix itself
Though, I can see `spectral_white_balance` above already spoils my ask ha ha

src/rawtoaces_util/image_converter.cpp
antond-weta: Please don't. Let them die already.

include/rawtoaces/image_converter.h
```diff
@@ -52,6 +52,13 @@ class ImageConverter
 
         enum class MatrixMethod
         {
+            /// Automatically choose the best available matrix method.
+            /// - If spectral sensitivity data for the camera is available,
+            ///   use `Spectral`.
+            /// - Otherwise, fall back to `Metadata`.
+            /// The "auto" mode solves this by first attempting "spectral", and if no
+            /// spectral data is available for the camera, it falls back to "metadata".
```
soswow: /nit last two lines are repeating 2 lines before that.

src/rawtoaces_util/image_converter.cpp
soswow: oh! I didn't know that's the direction we have. Sry, I don't have good feel for what's better, I thought it's short and neat. Let me kill the rest of them them, so it's not confusinb

mikaelsundell:
- The ACES group is moving away from using IDTs. Should we make rawtoaces consistent with this naming convention?
- rawtoaces_util_priv.h is now introduced. Should we include it in image_converter.cpp? If not, would it be better to move it to the tests? At the moment, we’re duplicating the function definition with a default argument for override_path.
- Running rawtoaces without -v produces no output. Should we at least print the file being processed or any failures as a default output in main.cpp

mikaelsundell:
"Yes, that is purely for unit-testing. No objection to moving to test. We should avoid duplicating things."

The reordering was required to remove the private header (which belongs in tests). It’s not actually needed in core, it was just there to satisfy the tests. Removing it exposed an ordering issue: functions must be declared before they are used.

... and Yes, it could of course be its own PR, it just depends on where we want to spend our time :-) The project is still manageable, but I’d like to move on as quickly as possible to focus on the GUI and the actual RAW processing in the project. Every PR/ change/ rebase is extra work to make sure it stays up to date with the changes in main.

@antond-weta Please double-check the updated comments for calculate_daylight_SPD and calculate_blackbody_SPD. I first wrote my own version and then refined it with copilot.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -40,14 +40,14 @@ struct CameraIdentifier
...
+ *
  * This function validates that the given path points to a regular file or symbolic link,
  * filters out unwanted files (system files like .DS_Store and certain image formats like EXR and JPG),
  * and adds valid file paths to the provided batch vector for further processing.
- * 
+ *
  * @param path The filesystem path to check
  * @param batch Reference to a vector of strings to add valid file paths to
- * @return true if the file was processed (either added to batch or filtered out), 
+ * @return true if the file was processed (either added to batch or filtered out),
```
antond-weta, line 50: this is probably a merge issue, the function has been changed to void.

mikaelsundell:
```
 * @return true if the file was processed (either added to batch or filtered out), 
 *         false if the file should be ignored
 */
void check_and_add_file(
    const std::filesystem::path &path, std::vector<std::string> &batch )
```

This is main I rebased from. This shouldn’t be fixed in this PR - the change will be made in a new one?

soswow: yeah. I've remove bool from return but forgot to update description

PR #192, opened 2025-09-10
More unit tests for util

soswow:


src/rawtoaces_util/image_converter.cpp
```diff
@@ -100,7 +99,8 @@ bool collect_image_files(
     }
     else
     {
-        check_and_add_file( path, batches[0] );
```
soswow, line 103: this was actually a bug cought by the unit test. Horray for unit tests.

src/rawtoaces_util/rawtoaces_util_priv.h
```diff
@@ -0,0 +1,27 @@
...
+
+// Contains the declarations of the private functions,
+// exposed here for unit-testing.
+
+namespace rta
+{
+namespace util
+{
+
+bool collect_image_files(
```
soswow: this is public API, so I guess I can drop it and use one from /include if you want

src/rawtoaces_util/image_converter.cpp
```diff
@@ -50,32 +50,31 @@ struct CameraIdentifier
  * @return true if the file was processed (either added to batch or filtered out), 
  *         false if the file should be ignored
  */
-bool check_and_add_file(
+void check_and_add_file(
```
soswow, line 53: boolean is not used in any way

src/rawtoaces_util/image_converter.cpp
antond-weta: what makes you think this was a bug? the code does exactly what  was intended, see main.cpp#40

src/rawtoaces_util/rawtoaces_util_priv.h
antond-weta: yes please, no need to duplicate things.

tests/test_image_converter.cpp
```diff
@@ -0,0 +1,481 @@
...
+        test_fix_metadata_float_make();
+        test_fix_metadata_destination_exists();
+        test_fix_metadata_source_missing();
+        test_fix_metadata_source_missing();
+        test_fix_metadata_unsupported_type();
+    }
+    catch ( const std::exception &e )
+    {
+        std::cerr << "Exception caught in main: " << e.what() << std::endl;
+        return 1;
```
antond-weta: not that this is important, but should this be `return unit_test_failures + 1;`, as returning 1 discards all other failures we've found so far

tests/test_image_converter.cpp
```diff
@@ -0,0 +1,481 @@
...
+    OIIO_CHECK_EQUAL( spec.find_attribute( "cameraModel" ), nullptr );
+}
+
+/// Tests fix_metadata with non-string, non-float attributes (should be ignored)
+void test_fix_metadata_unsupported_type()
+{
+    OIIO::ImageSpec spec;
+
+    // Add integer attribute (this should be ignored by fix_metadata)
+    spec["Make"] = 42; // Integer, not string or float
```
antond-weta: why do we think that camera make can be float, but not int

tests/test_image_converter.cpp
```diff
@@ -0,0 +1,481 @@
...
+    OIIO_CHECK_EQUAL( paths[1], "/preferred/path2" );
+#endif
+
+    // Clean up
+    unset_env_var( "RAWTOACES_DATA_PATH" );
+    unset_env_var( "AMPAS_DATA_PATH" );
+}
+
+/// Tests database_paths with Windows-style path separator
+void test_database_paths_windows_separator()
```
antond-weta: why do we test the windows separator, but no the unix one?

src/rawtoaces_util/image_converter.cpp
soswow: right. That is rather specific expectation. If it was private API and used internally, I'd see no problem with that. But expecting caller to know that expectation seems specific. Happy to revert and modify the test. You can imagine test is failing when `std::vector<std::vector<std::string>> batches;` is empty

antond-weta: I agree that the split between main/util is not optimal here, perhaps the outer loop (main.cpp#44) should be also moved inside `collect_image_files`. 
My point is, creating a new batch for each individual file was not the intended use. 
The logic was: `rawtoaces file1 file2 dir1 dir2 file3` should create 3 batches: [file1, file2, file3], [dir1/\*], [dir2/\*]

tests/test_image_converter.cpp
soswow: I don't know 🤷 That's the main problem not doing TDD and writing tests (contract) after impl. was written. (especially by a different brain)
```
if ( type.basetype == OIIO::TypeDesc::STRING )
    spec[dst_name] = src_attribute->get_string();
else if ( type.basetype == OIIO::TypeDesc::FLOAT )
    spec[dst_name] = src_attribute->get_float();
```
You tell me why we want attribute to be allows in as a float and a string, but not int. Happy to fix prod. code if there is a way to remove it.
This test just tests functionality that is already there.

antond-weta: Oh, I see what happened. The fix_metadata() originally had more attributes than just make and model. Could you delete that "else if float" branch for now, we'll add the allowed types to the mapping dictionary later when needed.

tests/test_image_converter.cpp
soswow: I don't need it. it's kinda stupid. If someone changes window separator, `test_database_paths_default` would brake.

src/rawtoaces_util/image_converter.cpp
soswow: you are right. I didn't know it was intended behaviour. Now I know. I've refactored basd on your suggestion. Now there is no need for boolean return, so batches is the return now. I've added a test that checks that exact behaviour.

include/rawtoaces/image_converter.h
```diff
@@ -11,16 +11,15 @@ namespace rta
...
-    const std::string &path, std::vector<std::vector<std::string>> &batches );
+/// Collect all files from given `paths` into batches.
+/// For each path that is a directory, create an entry in the returned batches and fill it with the file names
+/// (unless all files are filtered out in which case no batch is created).
+/// All paths that are files are added to the first batch.
+/// @param paths vector of paths to files or directories to process.
+/// @return vector of batches, where each batch contains files from one input path.
+///         Non-existent paths are skipped with an error message.
+std::vector<std::vector<std::string>>
+collect_image_files( const std::vector<std::string> &paths );
```
soswow, line 23: This is obv. a breaking change. You decide if it worth it. I think it does, considering we are still in beta of v2

src/rawtoaces_util/image_converter.cpp
```diff
@@ -50,60 +50,75 @@ struct CameraIdentifier
...
+    batches.erase(
+        std::remove_if(
+            batches.begin(),
+            batches.end(),
+            []( const std::vector<std::string> &batch ) {
+                return batch.empty();
+            } ),
+        batches.end() );
+
+    return batches;
```
soswow, line 111: I can see how it is used later on in `main`
```
for ( auto const &batch: batches )
    {
        for ( auto const &input_filename: batch )
        {
```
so, this ^ would deal with empty batches no problem, but from API perspective it doesn't make sense to have empty batch in return. So removing empty won't brake consumer, but make result cleaner.

src/rawtoaces_util/rawtoaces_util_priv.h
soswow: done

tests/test_image_converter.cpp
soswow: I see. makes sense. I am very big fan of removing code =)

include/rawtoaces/image_converter.h
antond-weta: cool

src/rawtoaces_util/image_converter.cpp
antond-weta: It does make sense though. It really depends on how the client needs the pipeline organised. 

Example: `rawtoaces dir1 dir2 dir3`, if it returns 2 batches, there is no way to know which one was empty on the client side if it is removed in lib.

soswow: ok. I've reverted it back. Empty batches are back.

PR #194, opened 2025-09-12
Add initial unit tests for  ImageConverter::parse_parameters

soswow:


tests/test_image_converter.cpp
```diff
@@ -522,6 +523,183 @@ void test_fix_metadata_unsupported_type()
...
+            result.output.substr( start_pos, end_pos - start_pos );
+        size_t pos = 0;
+        while ( ( pos = illuminant_section.find( "\n", pos ) ) !=
+                std::string::npos )
+        {
+            illuminant_count++;
+            pos += 1;
+        }
+        // Subtract 1 for the header line
+        illuminant_count = ( illuminant_count > 0 ) ? illuminant_count - 1 : 0;
```
soswow, line 754: Happy to drop that part. Makes test too fragile. If we decide to add some post-message after list of illuminants - it will blow up, which might not what we want. At the same time if we decide to add a new hardcoded illuminant, this test won't brake, even though we might want it to.

tests/materials/test-database/camera/Karamba_M2_380_780_5.json
```diff
@@ -0,0 +1,19 @@
+{
+    "header": {
+        "schema_version"         : "0.1.0",
+        "manufacturer"           : "karamba",
+        "model"                  : "M2",
+        "description"            : "data for unit tests",
+        "document_creator"       : "sasha",
+        "document_creation_date" : "2017-04-11T12:00:00Z",
+        "license"                : "Some-License-4.2"
```
antond-weta: I believe everything we put into this repo has to be under Apache-2.0, as per the project charter.

tests/materials/test-database/camera/Karamba_M2_380_780_5.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "spectral_data": {
+        "units"                 : "relative",
+        "index": {
+            "main": ["R", "G", "B"]
+        },
+        "data": {
+        }
+    }
+}
```
antond-weta: so why not remove even more unnecessary data and create the files on the fly?

tests/materials/test-database/illuminant/iso4242_380_780_5.json
```diff
@@ -0,0 +1,18 @@
+{
+    "header": {
+        "schema_version"         : "0.1.0",
+        "illuminant"             : "iso4242",
```
antond-weta: Quite interesting! 
> ISO 4242:1980
Cinematography — Recording head gaps for two sound records on 16 mm magnetic film — Positions and width dimensions

Maybe use a more generic name, unrelated to any standards?

antond-weta (approved):

PR #200, opened 2025-09-22
Refactor tests to actually run and not exit silently as well as add new one for data-dir parameter

soswow:


tests/test_image_converter.cpp
```diff
@@ -21,29 +21,92 @@
 
 using namespace rta::util;
 
+std::string convert_linux_path_to_windows_path( const std::string &path )
```
soswow, line 34: simplifies tests related to path parsing.

tests/test_image_converter.cpp
```diff
@@ -21,29 +21,92 @@
...
+    return OIIO::Strutil::join( segments, ";" );
+}
+
+/// Executes a rawtoaces command with the given arguments and captures its output.
+///
+/// @param args Vector of command-line arguments to pass to rawtoaces (excluding program name).
+///             For example, {"--list-cameras"} or {"--list-illuminants", "--verbose"}.
+///
+/// @return std::string containing the captured stdout output from the rawtoaces execution
+std::string run_rawtoaces_command( const std::vector<std::string> &args )
```
soswow, line 75: I've discovered that not all tests were running previouslu, because of the exit(0) inside `parse_parameters`.

tests/test_image_converter.cpp
```diff
@@ -21,29 +21,92 @@
...
+{
+    // Build the command line - from build/tests/ to build/src/rawtoaces/
+    std::string command = "../src/rawtoaces/rawtoaces";
```
soswow: [ ] TODO: deal with windows

tests/test_image_converter.cpp
```diff
@@ -392,14 +455,14 @@ void test_database_paths_default()
     OIIO_CHECK_EQUAL( paths.empty(), false );
 
 // On Unix systems, should have both new and legacy paths
-#if !defined( WIN32 ) && !defined( WIN64 )
-    OIIO_CHECK_EQUAL( paths.size(), 2 );
-    OIIO_CHECK_EQUAL( paths[0], "/usr/local/share/rawtoaces/data" );
-    OIIO_CHECK_EQUAL( paths[1], "/usr/local/include/rawtoaces/data" );
-#else
+#ifdef WIN32
```
soswow, line 494: simpler then `#if !defined( WIN32 ) && !defined( WIN64 )`

tests/test_image_converter.cpp
```diff
@@ -464,33 +510,78 @@ void test_database_paths_both_env()
...
+    OIIO_CHECK_EQUAL( paths[0], to_os_path( "/env/path1" ) );
+    OIIO_CHECK_EQUAL( paths[1], to_os_path( "/env/path2" ) );
 
     // Clean up
     unset_env_var( "RAWTOACES_DATA_PATH" );
     unset_env_var( "AMPAS_DATA_PATH" );
 }
 
+/// Tests convert_linux_path_to_windows_path utility function
+void test_convert_linux_path_to_windows_path()
```
soswow, line 606: a bit meta. testing testing utility. otherwise how do I know it works 🤷🏻‍♂️

tests/test_image_converter.cpp
```diff
@@ -654,41 +709,24 @@ void test_parse_parameters_list_cameras()
...
-    while ( ( pos = result.output.find( " / ", pos ) ) != std::string::npos )
-    {
-        camera_count++;
-        pos += 3; // Move past " / "
-    }
-    OIIO_CHECK_EQUAL( camera_count, 2 );
+        lines[0],
+        "Spectral sensitivity data is available for the following cameras:" );
+    OIIO_CHECK_EQUAL( lines[1], "Mamiya / Mamiya 7" );
+    OIIO_CHECK_EQUAL( lines[2], "Canon / EOS R6" );
```
soswow: much simpler assertion then confoluted previously

tests/test_image_converter.cpp
```diff
@@ -18,34 +18,134 @@
...
+#    include <io.h>
+#    include <process.h>
+#    include <direct.h> // for _getcwd
+#else
+#    include <sys/stat.h> // for mkfifo
+#    include <unistd.h>   // for getcwd
+#endif
+
+#include "../src/rawtoaces_util/rawtoaces_util_priv.h"
+#include <rawtoaces/image_converter.h>
```
soswow: I think it's merge conflict artifact. to be corrected.

tests/test_image_converter.cpp
```diff
@@ -18,34 +18,134 @@
...
+    {
+        output += buffer;
+    }
+
+    // Get exit status and validate
+    int exit_status = _pclose( pipe );
+    OIIO_CHECK_EQUAL( exit_status, 0 );
+#else
+    // Unix implementation using popen
+    FILE *pipe = popen( command.c_str(), "r" );
```
antond-weta: the code seems identical, can't we just set up a macro holding either `popen` or `_popen`; same for `pclose`

tests/test_image_converter.cpp
```diff
@@ -18,34 +18,134 @@
...
+
 // Cross-platform environment variable helpers
 /*
 Standard C Library vs POSIX
 getenv() - Part of standard C library (C89/C99) - available everywhere
 setenv()/unsetenv() - Part of POSIX standard - only on Unix-like systems
 */
 #ifdef WIN32
-void set_env_var( const char *name, const char *value )
+void set_env_var( std::string name, std::string value )
```
antond-weta: better to use const ref

tests/test_image_converter.cpp
antond-weta: I assume you've made sure that you don't need to check for WIN64

soswow: you assumed wrong. I believe it was just a crap AI code that thought that would be the way to say "If UNIX like OS" 🤷

tests/test_image_converter.cpp
```diff
@@ -18,34 +18,131 @@
...
+    std::string output;
+    char        buffer[4096];
+    while ( fgets( buffer, sizeof( buffer ), pipe ) != nullptr )
+    {
+        output += buffer;
+    }
+
+    // Get exit status and validate
+    int exit_status = platform_pclose( pipe );
+    OIIO_CHECK_EQUAL( exit_status, 0 );
```
lgritz, line 106: Just as an aside...

I'm glad you find OIIO's little unittest library helpful! 

We make the unittest.h header available in case it's of use to others (as it is here), but we don't really consider it a supported part of the OIIO public APIs, so I can't promise that we'll be careful to never do anything that might break your downstream usage.

I think it's fine for you to *copy* it and rename to R2A_CHECK_FOO so that you aren't directly dependent on the version you get from OIIO, and aren't brittle to any future changes made to these macros on the OIIO side. That also would allow you to customize them on your end to your particular taste or use cases.

That's the approach that OpenColorIO did many years ago.

PR #205, opened 2025-09-23
adds nanobind scaffolding for future python bindinds

antond-weta:
This adds all necessary code to start working on python bindings.
- configures nanobind
- adds a target `rawtoaces_bindings`
- adds some minimal bindings for testing purposes 

Note, I've only added minimal bindings necessary to convert an image. The stuff I've implemented:
- ImageConverter.process_image() - an example of bindings for a class and a method. It is working, and is enough to get an image converted with the default settings 
- ImageConverter.settings.WB_method - an example of bindings of a nested class and an enum.

I have made sure that the code builds on all platforms on CI, but I have only tested the python code on my machine (Mac).
Here is how I tested:
- go to the folder where the built bindings library is located, of add the path to `PYTHON_PATH` and/or `DYLD_LIBRARY_PATH`, this may be platform-dependent,
- execute this python code:
```
% python
>>> import rawtoaces
>>> converter = rawtoaces.ImageConverter()
>>> converter.process_image("/path/to/a/raw/file")
```

Next steps from here:

0. Set up python-based unit tests for future bindings. Ideally all bindings should have 100% test coverage.
1. Implement bindings for all properties and file-based methods from `rta::util::ImageConverter`, which do not use any OpenImageIO types as parameters or return values. The bindings in OIIO are currently implemented with pybind, but there is a plan to rewrite them with nanobind.
2. Implement bindings for `rta::core::SpectralSolver` and `rta::core::MetadataSolver`.
3. If the work to convert OIIO to nanobind hasn't been done yet, implement bindings for the `OIIO::ImageBuf`-based API by converting the ImageBufs to/from numpy arrays.
4. After OIIO has been converted to nanobind, finish up implementing the `OIIO::ImageBuf`-based API by exposing the OIIO types on the python side.


src/bindings/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+elseif ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+    target_compile_options ( nanobind-static    PRIVATE -Wno-error=zero-length-array )
+    target_compile_options ( rawtoaces_bindings PRIVATE -Wno-error=zero-length-array )
+endif()
+
+set_target_properties( rawtoaces_bindings PROPERTIES
+    OUTPUT_NAME "rawtoaces"
+    #EXPORT_NAME "rawtoaces"
+    #SOVERSION ${RAWTOACES_MAJOR_VERSION}.${RAWTOACES_MINOR_VERSION}.${RAWTOACES_PATCH_VERSION}
+    #VERSION ${RAWTOACES_VERSION}
```
soswow, line 31: comments. needed?

build_scripts/install_deps_windows.bash
```diff
@@ -5,4 +5,5 @@ set -ex
 vcpkg install \
     ceres:x64-windows \
     nlohmann-json:x64-windows \
-    openimageio:x64-windows
+    openimageio:x64-windows \
```
lgritz, line 8: I'm not sure this builds and installs the OIIO python bindings by default. You may need some kind of modifier.

src/bindings/CMakeLists.txt
antond-weta: I don't know. I think the binary should be versioned, but uncommenting that makes configure fail. 

Also I don't know if this whole 'export' thing is needed, I use it to change the library name, as the 'rawtoaces' target name is already taken. This also makes the library install with the rest of targets, I don't know if the destination should be different. This is good enough for now. We'll need somebody more familiar with python tools deployment to finalise this, I suppose.

build_scripts/install_deps_windows.bash
antond-weta: Noted. Will check. We don't need OIIO on the python side that right now.

PR #206, opened 2025-09-24
analysis

antond-weta:
## Description

Set up static analysis CI action. 

Was planning on using SonarCloud, but that seemed too complicated to me, so I ended up using Clang static analyser via CodeChecker.

This PR also fixes all existing SA warnings, so SA check should pass.

## Tests

No tests as this PR doesn't introduce any new code. It does change some code to fix warnings, mostly just variable renaming end explicit casting.

## Checklist:

  doesn't change the API.)
  nonconforming.


soswow:
nice

PR #212, opened 2025-10-19
Add more unit tests

soswow:
## Description

Adding more tests

## Checklist:

- [x I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/rawtoaces/blob/main/docs/CONTRIBUTING.md).
  doesn't change the API.)
  nonconforming.


src/bindings/CMakeLists.txt
```diff
@@ -20,8 +20,8 @@ if ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" )
         target_compile_options ( rawtoaces_bindings PRIVATE -Wno-error=zero-length-bounds )
     endif ()
 elseif ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
-    target_compile_options ( nanobind-static    PRIVATE -Wno-error=zero-length-array )
-    target_compile_options ( rawtoaces_bindings PRIVATE -Wno-error=zero-length-array )
+    target_compile_options ( nanobind-static    PRIVATE -Wno-error=zero-length-array -Wno-zero-length-array )
+    target_compile_options ( rawtoaces_bindings PRIVATE -Wno-error=zero-length-array -Wno-zero-length-array )
```
soswow, line 24: without this it was giving bunch of warning for stuff coming from python binding library headers

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -416,6 +416,7 @@ bool SpectralSolver::calculate_WB()
     {
         std::cerr << "ERROR: camera needs to be initialised prior to calling "
                   << "SpectralSolver::calculate_WB()" << std::endl;
+        // TODO: Should here be return false?
```
soswow: @antond-weta is that on purpose? If not, I'll make a change. All other similar places (There are two more we return false)

src/rawtoaces_util/image_converter.cpp
```diff
@@ -340,6 +340,7 @@ bool prepare_transform_spectral(
 
         if ( !success )
         {
+            // TODO: Potentially remove, since this code path is not reachable due to camera lookup success in the previous step.
```
soswow: calling @antond-weta to decide if I understand that right

tests/test_image_converter.cpp
```diff
@@ -800,6 +927,632 @@ void test_parse_parameters_list_illuminants( bool use_dir_path_arg = false )
     OIIO_CHECK_EQUAL( lines[3], "my-illuminant" );
 }
 
+/// Tests that prepare_transform_spectral fails when no camera manufacturer information is available (should fail)
+void test_missing_camera_manufacturer()
```
soswow, line 935: I was not able to change .dng we have such that it has no make 🤷 no matter what I did with exiftools, new .dng always has "DNG" as a make. Hence I couldn't make this test go through main gate like other ones.

src/bindings/CMakeLists.txt
antond-weta: could you try if suppressing the warnings locally around the nanobind includes in rawtoaces_bindings would work?

src/rawtoaces_core/rawtoaces_core.cpp
antond-weta: not on purpose, please fix. thanks.

src/rawtoaces_util/image_converter.cpp
antond-weta: please replace with `assert(success);`

build_scripts/vcpkg.json
```diff
@@ -3,7 +3,7 @@
     "dependencies": [
         "ceres",
         "nlohmann-json",
-        "openimageio",
+        { "name": "openimageio", "features": [ "libraw" ] },
```
soswow, line 6: In previous windows build dependency was brought in as:
```
openimageio:x64-windows@3.0.9.1#1
```
with this change it is now
```
openimageio[core,libraw]:x64-windows@3.0.9.1#1
```

soswow: Windows tests are green now.
Previously, when I was testing on local window machine the problem was that `oiio` I had didn't have libraw plugin enabled when I was trying to do convertion as part of a test. So, this most likely what fixed it.

cmake/modules/Findlibraw.cmake
soswow, line 1: My LLM convinced that this change is required. Though I am not 100% understrand what's going on here and I Am not convinced it is required. I feel like `vcpkg.json` change is where the problem was.

soswow:
I had this realisation - all this stuff I did with running executable directly via _open (which is good from end-to-end persepective) will not count into coverage counter =\ most likely

src/bindings/CMakeLists.txt
soswow: I don't know what that means, tbh. Can you expand? Don't forget, not a native c++'er here

antond-weta: ignore that, it is minor. I may do that later separately

tests/test_image_converter.cpp
```diff
@@ -800,6 +927,593 @@ void test_parse_parameters_list_illuminants( bool use_dir_path_arg = false )
...
+
+    // Create observer data (so observer data loading succeeds)
+    test_dir.create_test_data_file( "cmf", { { "illuminant", "D65" } } );
+
+    // Create illuminant data (so illuminant data loading succeeds)
+    test_dir.create_test_data_file( "illuminant", { { "illuminant", "D65" } } );
+
+    // Use the DNG file with camera make but no model
+    std::string test_file =
+        "../../tests/materials/blackmagic_cinema_camera_cinemadng_no_model.dng";
```
antond-weta: Adding a 5mb image file for testing just this small use case seems a bit excessive to me. I would suggest doing this: let's skip this part in the end-to-end tests. We have a `configure()` method which takes an ImageSpec instead of a file path. We can try using that to modify the ImageSpec for testing purposes: load a file into an ImageBuf, grab its ImageSpec, delete the camera model atribute, feed to `configure()`. This way we could test all sorts of bad metadata use cases.

antond-weta (approved):
LGTM

PR #213, opened 2025-10-31
add native rawtoaces CI runners

antond-weta:
## Description

Add native CI runners (aswf/ci-rawtoaces) for VFX year 2024, 2025, 2026.

## Tests

No changes to the code

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -80,7 +80,7 @@ jobs:
         shell: bash
         run: |
           conan profile detect --force
-          conan install . --output-folder=build_deps --build=missing
+          conan install --requires=nlohmann_json/3.12.0 --requires=nanobind/2.9.2 --generator CMakeDeps --generator CMakeToolchain --output-folder=build_deps --build=missing
           python -m pip install pytest
     
       - name: Configure CMake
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +41,14 @@ jobs:
           - vfxyear: 2023
             container: aswf/ci-osl:2023-clang15
           - vfxyear: 2024
-            container: aswf/ci-osl:2024-clang17
+            container: aswf/ci-rawtoaces:2024.4
+            toolchain_file: "/__w/rawtoaces/rawtoaces/build_deps/conan_toolchain.cmake"
```
soswow, line 45: these seem to be absolute and rather magical paths. are we sure in those?

soswow (approved):

.github/workflows/ci.yml
antond-weta: That's where we install the dependencies to in a preceding step `conan install --output-folder=build_deps`
That step shouldn't be necessary at all, as the rawtoaces-specific containers come with all deps preinstalled. Hopefully we'll resolve the issues soon and remove this.

PR #215, opened 2025-11-21
a few small tweaks to the public API

antond-weta:
## Description

Small tweaks to the public API to squeeze in before 2.0.
The changes are really minor:
- add a `const` qualifier to a few methods
- make the reference to hints constant in configure()

## Tests

no new code introduced

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/include/rawtoaces/image_converter.h
+++ b/include/rawtoaces/image_converter.h
@@ -147,11 +147,11 @@ public:
     bool parse_parameters( const OIIO::ArgParse &arg_parser );
 
     /// Collects all illuminants supported by this version.
-    std::vector<std::string> supported_illuminants();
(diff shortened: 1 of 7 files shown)


include/rawtoaces/image_converter.h
```diff
@@ -147,11 +147,11 @@ class ImageConverter
     bool parse_parameters( const OIIO::ArgParse &arg_parser );
 
     /// Collects all illuminants supported by this version.
-    std::vector<std::string> supported_illuminants();
+    std::vector<std::string> get_supported_illuminants() const;
```
soswow, line 150: love it. method should be verbs

PR #219, opened 2025-11-26
[fix] remove breaking const qualifier on hints in configure

antond-weta:
## Description

The const qualifier on hints added in [215](https://github.com/AcademySoftwareFoundation/rawtoaces/pull/215) breaks conversion. Rolling that change back.

## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/rawtoaces/blob/main/docs/CONTRIBUTING.md).
  doesn't change the API.)
  nonconforming.


soswow (approved):
I usually prefer to have regression test to accompony the revert/bug-fix. But it's ok to merge first and think of testing later.

PR #223, opened 2025-12-10
[python] more ImageConverter bindings

antond-weta:
## Description

This contains the python bindings for the rest of the ImageConverter functionality, except of the ones requiring OIIO objects.
OIIO doesn't currently have nanobind bindings, we'll add those after OIIO switches from pybind to nanobind.

## Tests

There are unit tests for all added bindings.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/include/rawtoaces/image_converter.h
+++ b/include/rawtoaces/image_converter.h
@@ -25,10 +25,10 @@ collect_image_files( const std::vector<std::string> &paths );
 class ImageConverter
 {
 public:
+    /// The structure containing all parameters needed to configure image conversion.
     struct Settings
     {
-        /// The  white balancing method to use for conversion can be specified
-        ///
+        /// The enumerator containing all supported white-balancing methods.
         enum class WBMethod
         {
             /// Use the metadata provided in the image file. This mode is mostly
@@ -48,8 +48,12 @@ public:
             /// the white balancing coefficients are calculated by an external
             /// tool.
             Custom
-        } WB_method = WBMethod::Metadata;
+        };
 
+        /// The selected white-balancing method to use for conversion.
+        WBMethod WB_method = WBMethod::Metadata;
+
+        /// The enumerator containing all supported colour transform matrix calculation methods.
         enum class MatrixMethod
         {
             /// Automatically choose the best available matrix method.
@@ -72,9 +76,12 @@ public:
             /// Specify a custom matrix in `colourMatrix`. This mode is useful if
             /// the matrix is calculated by an external tool.
             Custom
-        } matrix_method = MatrixMethod::Auto;
+        };
 
-        /// Cropping mode.
+        /// The selected colour transform matrix calculation method to use for conversion.
+        MatrixMethod matrix_method = MatrixMethod::Auto;
+
+        /// The enumerator containing all supported cropping modes.
         enum class CropMode
         {
             /// Write out full sensor area.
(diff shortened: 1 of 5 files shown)


src/bindings/py_util.cpp
```diff
@@ -3,37 +3,193 @@
...
+    settings.def_prop_rw(
+        "crop_box",
+        []( const ImageConverter::Settings &settings ) {
+            std::vector<int> result( 4 );
+            for ( size_t i = 0; i < 4; i++ )
+                result[i] = settings.crop_box[i];
+            return result;
+        },
+        []( ImageConverter::Settings &settings, const std::vector<int> &box ) {
+            if ( box.size() == 4 )
```
soswow, line 168: what happens if user doesn't provide array of 4?

tests/python/test_image_converter.py
```diff
@@ -62,6 +103,120 @@ def test_process_image_with_invalid_path(self):
...
+        assert converter.configure(path)
+        wb = converter.get_WB_multipliers()
+        assert len(wb) == 4
+        assert abs(wb[0] - 1.58982515335083) < 0.0001
+        assert abs(wb[1] - 1) < 0.0001
+        assert abs(wb[2] - 1.2651821374893188) < 0.0001
+        assert abs(wb[3] - 0) < 0.0001
+
+        print("get_IDT_matrix", converter.get_IDT_matrix())
+        print("get_CAT_matrix", converter.get_CAT_matrix())
```
soswow: are these meant to stay?

tests/python/test_image_converter.py
```diff
@@ -62,6 +103,120 @@ def test_process_image_with_invalid_path(self):
...
+
+    def test_converter_get_WB_multipliers(self):
+        """Test uninitialised ImageConverter returns empty WB multipliers"""
+        import os
+        converter = rawtoaces.ImageConverter()
+        try:
+            wb = converter.get_WB_multipliers()
+            assert len(wb) == 0
+        except Exception:
+            pytest.fail(Exception)
```
soswow: Should it be:
```
except Exception as e:
    pytest.fail(f"Unexpected exception: {e}")
```
?

src/bindings/py_util.cpp
antond-weta: added an exception

tests/python/test_image_converter.py
antond-weta: fixed

tests/python/test_image_converter.py
antond-weta: fixed

include/rawtoaces/image_converter.h
```diff
@@ -94,39 +104,106 @@ class ImageConverter
         /// folder.
         std::string illuminant;
 
-        float headroom            = 6.0;
-        int   WB_box[4]           = { 0 };
-        float custom_WB[4]        = { 1.0, 1.0, 1.0, 1.0 };
+        /// Highlight headroom factor.
```
soswow, line 107: I appreciate all the new comments!

include/rawtoaces/image_converter.h
```diff
@@ -94,39 +104,106 @@ class ImageConverter
...
+        float headroom = 6.0;
+
+        /// Box to use for white balancing when `WB_method` == `WBMethod::Box`.
+        /// (default = (0,0,0,0) - full image)
+        int WB_box[4] = { 0, 0, 0, 0 };
```
soswow, line 112: Interesting. Previously `int   WB_box[4]           = { 0 };` - was/is there some code somewhere that does if/else for case when it's an array with one element. If we dont' do it any more - maybe it can be culled?

antond-weta: The array is still 4 elements, `= {0}` is a short form of zero initialisation and is equivalent to ` = {0, 0, 0, 0}`. 
https://www.en.cppreference.com/w/c/language/array_initialization.html

I just expanded it to match the comment above.

PR #225, opened 2025-12-12
Do not install deps when cache hit

soswow:
do not install windows deps


antond-weta:
This looks kludgy. Vcpkg should not reinstall what is already present. Can we try caching the `binary cache` directory instead of the `C:/vcpkg/installed` 
https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-local?pivots=shell-bash

Alternatively we could explore switching from vcpkg to conan.

PR #226, opened 2025-12-13
Complete edge cases main.cpp coverage

soswow:
more tests

### DIFF
--- a/src/rawtoaces/main.cpp
+++ b/src/rawtoaces/main.cpp
@@ -34,8 +34,8 @@ int main( int argc, const char *argv[] )
     }
 
     // Gather all the raw images from arg list
-    std::vector<std::vector<std::string>> batches;
-    batches = rta::util::collect_image_files( files );
+    std::vector<std::vector<std::string>> batches =
+        rta::util::collect_image_files( files ); // LCOV_EXCL_LINE
 
     // Process raw files
     bool empty  = true;


src/rawtoaces/main.cpp
```diff
@@ -19,10 +19,7 @@ int main( int argc, const char *argv[] )
     arg_parser.arg( "filename" ).action( OIIO::ArgParse::append() ).hidden();
     converter.init_parser( arg_parser );
 
-    if ( arg_parser.parse_args( argc, argv ) < 0 )
-    {
-        return 1;
```
soswow, line 24: This code is unreachable. `m_exit_on_error` inside parser controls behaviour. It is `true` by default , which means it will exit instead of returning negative value.

src/rawtoaces/main.cpp
```diff
@@ -37,8 +34,8 @@ int main( int argc, const char *argv[] )
     }
 
     // Gather all the raw images from arg list
-    std::vector<std::vector<std::string>> batches =
-        rta::util::collect_image_files( files );
```
soswow, line 41: this change is purely for coverage purposes. 
Without it current setup thinks this line is not visited (with lines above and below it is). Something to do with AppleLang and lcov way with multi-line

Other options:
* Add `LCOV_EXCL_LINE` comment on this line, and it will be excluded which is ok
* Switch to Clang's source-based coverage (-fprofile-instr-generate -fcoverage-mapping) instead of gcov (a lot of work)
* Have non 100% coverage

antond-weta: Interesting. Would `auto batches` work there? If not, please leave a comment explaining why the lines were split, as anybody looking at that code would have the urge to combine them back.

soswow: I think I've tried that before as well. 
image

doesn't work =( 
If we leaving a comment, I rather then add `// LCOV_EXCL_LINE - to cover the bug in coverage tool.`

PR #227, opened 2025-12-14
More tests 3

soswow:
moar tests


tests/test_image_converter.cpp
```diff
@@ -402,6 +403,46 @@ class TestDirectory
     std::string database_dir;
 };
 
+void assert_success_conversion( const std::string &output )
```
soswow, line 406: just moving it from down below

antond-weta (approved):
LGTM

PR #230, opened 2025-12-25
[analysis] fix warnings

antond-weta:
## Description

The static analyser was showing warnings about four matrices. Those were constructed as global static vectors. Being a global object, the constructor gets called during the app initialisation, not client code, so in case of an exception in the constructor it would be impossible to catch. Since each of the matrices being used in exactly one place, I just moved them closer to the code using them, so the static objects getting initialised in the client code.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -8,9 +8,9 @@ permissions:
 on:
   schedule:
     # Run nightly
-    - cron: "0 8 * * *"
+#    - cron: "0 8 * * *"
     # Run weekly
-    # - cron: "0 0 * * 0"
+     - cron: "0 0 * * 0"
   push:
     # Run on pushes only to main or if the branch name contains "analysis"
     branches:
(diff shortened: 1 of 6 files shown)


include/rawtoaces/rawtoaces_core.h
```diff
@@ -12,18 +12,24 @@ namespace core
 
 // clang-format off
 
+// NOLINTBEGIN
+/// Deprecated, not currently used and will be removed in v3.0.
```
antond-weta, line 16: Since this is a public API we can't just delete the matrix. I couldn't also use the `[[deprecated]]` attribute, as the constructor being always invoked on the app start up, even if we never use the matrix, and we have error on warnings enabled. So I'm just suppressing the analyser warning there and commenting that the matrix should not be used.

.github/workflows/analysis.yml
```diff
@@ -8,9 +8,9 @@ permissions:
 on:
   schedule:
     # Run nightly
-    - cron: "0 8 * * *"
+#    - cron: "0 8 * * *"
     # Run weekly
-    # - cron: "0 0 * * 0"
+     - cron: "0 0 * * 0"
```
antond-weta, line 13: we don't have any outstanding analyser errors, so weekly should be good enough.

PR #231, opened 2026-01-03
Use OIIO RAW extensions for file validation

mikaelsundell:
Skip hard-coding file extensions when collecting input files and instead query OIIO for the RAW extensions it actually supports. The list is initialized once and reused both for directory scanning and --list-formats.

This also prepares for GUI use, where the same extension list will be needed for file dialogs and drag-and-drop filtering.

No API changes, no tests added at this point.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_util/image_converter.cpp
+++ b/src/rawtoaces_util/image_converter.cpp
@@ -81,9 +81,7 @@ parse_raw_extensions( const std::string &extensionlist )
 /**
  * Returns the set of file extensions supported for RAW image processing.
  *
- * The extensions are discovered from OpenImageIO at runtime, filtered to the
- * "raw" format group, normalized to lowercase, and prefixed with a dot.
- * The result is initialized once and reused.
+ * The extensions are discovered from OpenImageIO at runtime.
  *
  * This function is the single source of truth for RAW format support and is
  * used both for input file filtering and for the "--list-formats" option.
@@ -93,7 +91,9 @@ parse_raw_extensions( const std::string &extensionlist )
 const std::set<std::string> supported_raw_extensions()
 {
     std::string extensionlist;
-    OIIO::getattribute( "extension_list", extensionlist );
+    bool        ok = OIIO::getattribute( "extension_list", extensionlist );
+    OIIO_ASSERT( ok && "OIIO did not provide extension_list" );
+
     return parse_raw_extensions( extensionlist );
 }
 
--- a/tests/test_image_converter.cpp
+++ b/tests/test_image_converter.cpp
@@ -397,15 +397,6 @@ void test_collect_image_files_mixed_valid_invalid_paths()
     OIIO_CHECK_EQUAL( batches[1].size(), 1 );
 }
 
-/// Tests supported_raw_extensions() under normal OIIO runtime conditions
-void test_supported_raw_extensions()
-{
-    std::cout << std::endl << "test_supported_raw_extensions()" << std::endl;
-
-    const auto &exts = rta::util::supported_raw_extensions();
-    OIIO_CHECK_EQUAL( exts.empty(), false );
-}
-
 /// Tests parsing of RAW extensions from a mixed OIIO extension list
 void test_parse_raw_extensions()
 {
@@ -1348,9 +1339,9 @@ void test_find_illuminant_camera_no_main_key()
     // Should fail because camera is not initialized (no "main" key)
     OIIO_CHECK_ASSERT( !success );
     OIIO_CHECK_ASSERT(
-        output.find(
-            "ERROR: camera needs to be initialised prior to calling "
-            "SpectralSolver::find_illuminant()" ) != std::string::npos );
+        output.find( "ERROR: camera needs to be initialised prior to calling "
+                     "SpectralSolver::find_illuminant()" ) !=
+        std::string::npos );
 }
 
 /// Tests that find_illuminant fails when camera data has "main" but with wrong size
@@ -1891,11 +1882,11 @@ void test_spectral_conversion_external_legacy_illuminant_success()
 
     // Create test directory with database
     TestFixture fixture;
-    auto       &test_dir = fixture.with_camera( "Blackmagic", "Cinema Camera" )
-                         .with_illuminant_custom(
-                             { { "schema_version", "0.1.0" },
-                               { "illuminant", "test_illuminant" } } )
-                         .build();
+    auto       &test_dir =
+        fixture.with_camera( "Blackmagic", "Cinema Camera" )
+            .with_illuminant_custom( { { "schema_version", "0.1.0" },
+                                       { "illuminant", "test_illuminant" } } )
+            .build();
 
     // Build command
     auto args = CommandBuilder()
@@ -2165,7 +2156,6 @@ int main( int, char ** )
         test_collect_image_files_mixed_valid_invalid_paths();
 
         // Tests for raw extensions
-        test_supported_raw_extensions();
         test_parse_raw_extensions();
 
         // Tests for database_paths


mikaelsundell:
You’re right - this does introduce an API change in the strict sense, as it adds a new public member and a new command-line option. What I meant was that the change is non-breaking: existing behavior and existing calls remain unaffected.

The new command-line option is informational only and simply exposes the list of RAW formats already supported internally via OIIO; it does not alter processing behavior or defaults.

The tricky part is whether we want to keep the new extension check in `check_and_add_file`, as this could potentially affect users who deliberately force formats that are not listed by OIIO.

I don’t mind adding a test to verify the reported formats if we agree this is the right direction. If the change in check_and_add_file feels too aggressive, we can either revert back to existing code or limit this logic to the GUI application instead.

mikaelsundell:
Perfect, will freeze the code as-is and add the last changes.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -38,12 +38,65 @@ struct CameraIdentifier
...
+
+            // OIIO Formats: "raw:cr2,nef,...;jpeg:jpg,..."
+            std::vector<std::string> groups;
+            OIIO::Strutil::split( extensionlist, groups, ";" );
+
+            for ( const auto &group: groups )
+            {
+                auto colon = group.find( ':' );
+                if ( colon == std::string::npos )
+                    continue;
```
soswow: Any way we get have a unit test that will get over here and over at line 61?

soswow:
There is `tests/python/test_image_converter.py` full of tests for python binding. like `test_converter_get_supported_cameras` that is already there

src/rawtoaces_util/image_converter.cpp
mikaelsundell: The continue statements are filters used while parsing OIIO’s extension_list, which is external input. They are internal implementation details.

The intended behavior of this function is to return the supported RAW extensions. By asserting the presence of well-known RAW formats such as .cr2 and .dng, the tests already verify that the parsing and filtering behave correctly. If any of the continue logic were incorrect, those extensions would be missing or incorrect and the test would fail.

That said, you still think an additional test is needed?

antond-weta: Thanks for adding the tests! Ideally we should be having close to 100% coverage over all newly added code. In some cases that would require splitting some functionality into a separate function, just so it can be executed will different inputs from unit tests. 

We need to get over 90% total coverage to get the OpenSSF Best Practices badge. Sasha has been working on improving the test coverage of the existing code, but ideally everybody adding new code should be doing their part as well.

lgritz: Just to temper expectations:

I think the three badge levels are graduated: no specific coverage requirement for "passing", 80% for silver, 90% for gold.

Every code base is different, but anecdotally, when we first started looking at improving OIIO's coverage, it was around 65% (as a sprawling, 10+ year old code base). It was easy (i.e. just a lot of hard work) to get to around 75%, since there were lots of HUGE uncovered sections that represented features we had that just weren't tested in the testsuite. But then going from 75% to 80+ was getting really hard -- it seemed like real work to eke out every additional 0.1%. We are currently at around 82% or so, and there is no low-hanging fruit left. Most of the uncovered lines now are in short isolated segments (sometimes just a few lines), and it's prohibitive to construct new tests that only add coverage to tiny portions of the code base each. While we're always striving to improve testing (and definitely aiming to get close to 100% coverage on NEW functionality), I consider 90% an essentially unattainable target for OIIO. Given the amount of work it takes to add coverage of each new block of uncovered lines, I think we could direct all development resources to test construction for literally years to get to > 90%. It would be possible, but not justifiable versus other things we could be working on.

antond-weta: Thanks for the insight, Larry. That is great info. OIIO is indeed huge and a pretty old code base. On the other hand, rawtoaces is pretty small, and a huge chunk of the code base has been rewritten from scratch about 6 months ago. So I think we are in a good position to get good coverage. I'm not sure if 90% is attainable, but we are currently sitting at 85%, and I can't say that we had spent huge amount of time to achieve that (Sasha may want to disagree).

I'm expecting the coverage to drop somewhat when we start working on the GUI component though.

soswow: yea, nah, it wasn't this hard. All branches of the code is pretty reachable. We are not at 90 or 95 only because I haven't gotten there yet. I am confident we can do it.

soswow:
@mikaelsundell I hope you can forgive me, but there was a refactoring-of-sort PR just merged that introduced some utils for tests (for setup and assertion). You might get a conflict and you might want to use those new goodies. sry about that, I was working on it for some time.

src/rawtoaces_util/image_converter.cpp
lgritz: Ah, very good. If you're already in high 80s and think adding more coverage isn't hard, by all means go for it. I personally believe that untested code should be presumed to be buggy. I've been humbled by my own bugs far too many times to trust myself when I don't have a test.

I'm just saying: It can be extremely hard to get close to 100%, at some point you get diminishing returns where the resources needed to write the tests are so great that they are not giving you a positive ROI versus taking a chance and fixing a bug if it is ever encountered. Don't beat yourself up about making that judgment call when you feel like you have to make it. At the end of the day, it's about using your finite development resources where they do the most good. Very often, the best use of resources is getting better test coverage. But sometimes there are bigger fish to fry. This is a matter of the project leader setting the priorities that feel right for the particular project.

tests/test_image_converter.cpp
```diff
@@ -397,6 +397,41 @@ void test_collect_image_files_mixed_valid_invalid_paths()
...
 }
 
+/// Tests supported_raw_extensions() under normal OIIO runtime conditions
+void test_supported_raw_extensions()
+{
+    std::cout << std::endl << "test_supported_raw_extensions()" << std::endl;
+
+    const auto &exts = rta::util::supported_raw_extensions();
+    OIIO_CHECK_EQUAL( exts.empty(), false );
+}
```
soswow: not sure if this test is needed.

tests/test_image_converter.cpp
```diff
@@ -648,6 +683,35 @@ std::string run_rawtoaces_with_data_dir(
...
+    {
+        if ( line == ".cr2" )
+            found_cr2 = true;
+        if ( line == ".dng" )
+            found_dng = true;
+    }
+
+    OIIO_CHECK_EQUAL( found_cr2, true );
+    OIIO_CHECK_EQUAL( found_dng, true );
+}
```
soswow, line 708: can we please have a negative case:

```
bool found_png = false;
...
        if ( line == ".png" )
            found_png = true;
...
OIIO_CHECK_EQUAL( found_png, false );
```

src/rawtoaces_util/image_converter.cpp
```diff
@@ -38,12 +38,77 @@ struct CameraIdentifier
...
+ * used both for input file filtering and for the "--list-formats" option.
+ *
+ * @return A constant reference to the set of supported RAW file extensions.
+ */
+const std::set<std::string> &supported_raw_extensions()
+{
+    static const std::set<std::string> extensions = [] {
+        std::string extensionlist;
+        if ( !OIIO::getattribute( "extension_list", extensionlist ) )
+            return std::set<std::string>{};
```
soswow: I believe these two lines (check) can be removed safely?
I am right that the result will be the same?
Or maybe we can assign default (empty vector) value and just call getattribute. If value is not found - it will just stay empty. no?

mikaelsundell: extension_list should exist for all valid OIIO builds,  now changed to:
```
    bool        ok = OIIO::getattribute( "extension_list", extensionlist );
    OIIO_ASSERT( ok && "OIIO did not provide extension_list" );
```

tests/test_image_converter.cpp
mikaelsundell: Yes, done!

tests/test_image_converter.cpp
mikaelsundell: No longer needed, removed!

soswow (approved):

PR #233, opened 2026-01-05
dynamic analysis

antond-weta:
## Description

This adds cmake config and CI actions to run dynamic analysis. 
Currently only the address and thread sanitisers are configured to run on CI, although the memory sanitiser can be run locally (I believe the OpenSSF badge only requires these three).

Currently the dynamic analysis gets run on the valid permutations of the WB and matrix methods (tests/run_valid.sh), I will add a script generating the launch commands for various permutations of invalid inputs later as a separate PR. 

## Tests

These are tests.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -104,7 +104,6 @@ jobs:
           -D RTA_CENTOS7_CERES_HACK=${{ inputs.ceres_hack }}
           -D CMAKE_INSTALL_PREFIX=${{ github.workspace }}/install
           -D CMAKE_CXX_STANDARD=17
-          -D CMAKE_BUILD_TYPE=${{ inputs.build_type }}
           -D CMAKE_TOOLCHAIN_FILE="${{ inputs.toolchain_file }}"
           -D CMAKE_BUILD_TYPE="${{ inputs.build_type }}"
           -D ENABLE_SHARED="${{ inputs.build_shared_libs }}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,8 @@ if ( NOT "${RTA_SANITISER_MODE}" STREQUAL "none" )
     elseif ( "${RTA_SANITISER_MODE}" STREQUAL "thread" )
         set ( RTA_SANITISER_COMPILE_OPTIONS "-fsanitize=thread" "-g" "-O1" )
         set ( RTA_SANITISER_LINK_OPTIONS "-fsanitize=thread" "-g" )
+    else ()
+        message ( FATAL_ERROR "Invalid RTA_SANITISER_MODE: ${RTA_SANITISER_MODE}" )
     endif ()
     
     target_compile_options( ${RAWTOACES_CORE_LIB} PUBLIC


.github/workflows/build-steps.yml
```diff
@@ -103,24 +106,31 @@ jobs:
           -D CMAKE_CXX_STANDARD=17
           -D CMAKE_BUILD_TYPE=${{ inputs.build_type }}
           -D CMAKE_TOOLCHAIN_FILE="${{ inputs.toolchain_file }}"
+          -D CMAKE_BUILD_TYPE="${{ inputs.build_type }}"
```
soswow, line 108: it is already there on line 107

.github/workflows/build-steps.yml
```diff
@@ -145,3 +155,9 @@ jobs:
           --rerun-failed 
           --output-on-failure 
           --test-dir build_config_test
+
+      - name: Run sanitiser checks
+        if: inputs.sanitiser_mode != 'none'
+        working-directory: build/src/rawtoaces
```
soswow, line 160: inside script (run_valid.sh) it says `# Execute from the build directory of the CLI tool` but we run it from a different place here.

CMakeLists.txt
```diff
@@ -111,6 +112,32 @@ if( BUILD_DOCS )
...
+    if ( "${RTA_SANITISER_MODE}" STREQUAL "address" )
+        set ( RTA_SANITISER_COMPILE_OPTIONS "-fsanitize=address" "-fno-omit-frame-pointer" "-g" "-O1" )
+        set ( RTA_SANITISER_LINK_OPTIONS "-fsanitize=address" "-g" )
+    elseif ( "${RTA_SANITISER_MODE}" STREQUAL "memory" )
+        set ( RTA_SANITISER_COMPILE_OPTIONS "-fsanitize=memory" "-fno-omit-frame-pointer" "-g" "-O2" "-fsanitize-memory-track-origins" )
+        set ( RTA_SANITISER_LINK_OPTIONS "-fsanitize=memory" "-g" "-fsanitize-memory-track-origins" )
+    elseif ( "${RTA_SANITISER_MODE}" STREQUAL "thread" )
+        set ( RTA_SANITISER_COMPILE_OPTIONS "-fsanitize=thread" "-g" "-O1" )
+        set ( RTA_SANITISER_LINK_OPTIONS "-fsanitize=thread" "-g" )
+    endif ()
```
soswow, line 133: do we care for `else` here with smth like 
`message(FATAL_ERROR "Invalid RTA_SANITISER_MODE: ${RTA_SANITISER_MODE}")` ?

.github/workflows/build-steps.yml
antond-weta: This is the built binary location, assuming 'build' is the build folder. Is it different for you?

.github/workflows/build-steps.yml
antond-weta: I guess that didn't work without quotes. Fixed.

CMakeLists.txt
antond-weta: done

soswow:
send it!

PR #235, opened 2026-01-06
Make src/rawtoaces_core/ be covered 100%

soswow:
Couple tests for `src/rawtoaces_core/rawtoaces_core.cpp` lines 706-738
## Checklist:

  doesn't change the API.)
  nonconforming.


src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1161,7 +1161,7 @@ void get_camera_XYZ_matrix_and_white_point(
     assert( sumVector( out_camera_XYZ_white_point ) != 0 );
 
     return;
-}
```
antond-weta, line 1164: perhaps that line is unreachable because of the `return`, can we try removing it?

tests/CMakeLists.txt
```diff
@@ -104,6 +105,7 @@ add_test ( NAME Test_Illum COMMAND Test_Illum )
 add_executable (
 	Test_DNGIdt
 	testDNGIdt.cpp
+	test_utils.cpp
```
antond-weta, line 108: Not entirely happy with exposing the rawtoaces_util code in the rawtoaces_core tests as an implicit dependency.
Hopefully we'll redo the logging soon, so the stderr capturing won't be needed anymore.

antond-weta (commented):
looks good

src/rawtoaces_core/rawtoaces_core.cpp
soswow: yeap. that was it.

tests/CMakeLists.txt
soswow: hopefully I got what you mean and did what asked.

PR #236, opened 2026-01-11
[feature] transform cache

antond-weta:
## Description

Implements colour transform caches. 
Adds a new `--disable-cache` command line parameter.

## Tests

There are unit tests for all new functionality.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/tests/test_image_converter.cpp
+++ b/tests/test_image_converter.cpp
@@ -2158,88 +2158,88 @@ int main( int, char ** )
 {
     try
     {
-        //        // Tests for collect_image_files
-        //        test_collect_image_files_directory();
-        //        test_collect_image_files_single_file();
-        //        test_collect_image_files_nonexistent_path();
-        //        test_collect_image_files_empty_directory();
-        //        test_collect_image_files_directory_with_only_filtered_files();
-        //        test_collect_image_files_multiple_paths();
-        //        test_collect_image_files_mixed_valid_invalid_paths();
-        //
-        //        // Tests for raw extensions
-        //        test_parse_raw_extensions();
-        //
-        //        // Tests for database_paths
-        //        test_database_paths_default();
-        //        test_database_paths_rawtoaces_env();
-        //        test_database_paths_ampas_env();
-        //        test_database_paths_both_env();
-        //        test_database_paths_override_path();
-        //
-        //        // Tests for utility functions
-        //        test_convert_linux_path_to_windows_path();
-        //
-        //        // Tests for fix_metadata
-        //        test_fix_metadata_both_attributes();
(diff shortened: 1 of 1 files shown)


src/rawtoaces_util/image_converter.cpp
```diff
@@ -394,67 +347,49 @@ bool prepare_transform_spectral(
...
-            std::cerr << "  ";
-            for ( auto &col: row )
-            {
-                std::cerr << col << " ";
-            }
-            std::cerr << std::endl;
-        }
-    }
+    // Calculate Input Device Transform (IDT) matrix.
+    success = fetch_matrix_from_illuminant(
```
soswow, line 385: convertion result is not used. (when false, we still return `true`)

src/rawtoaces_util/transform_cache.h
```diff
@@ -0,0 +1,56 @@
...
+    WB_from_illuminant_cache;
+
+// -----------------------------------------------------------------------------
+// Illuminant from WB cache
+// -----------------------------------------------------------------------------
+
+using CameraAndWBDescriptor = std::tuple<
+    std::string,          // camera make
+    std::string,          // camera model
+    std::array<double, 3> // white balancing weights
```
soswow, line 45: fetching mechanism doesn't seem to have tolerance built in, right? So when we want to pick camera by WB values exact double will be used. Is that desired/as expected?

tests/testCache.cpp
```diff
@@ -0,0 +1,233 @@
...
+    output = capture_stderr( [&]() {
+        std::tuple<std::string, std::string, std::array<double, 3>> tuple = {
+            "a", "b", { 1.1, 2.2, 3.3 }
+        };
+        std::cerr << tuple << std::endl;
+    } );
+    ASSERT_CONTAINS( output, "a, b, (1.1, 2.2, 3.3)" );
+}
+
+void testCache_transform_caches()
```
soswow, line 250: not called in `main()` so test is ignored. not sure if intentional

src/rawtoaces_util/image_converter.cpp
antond-weta: fixed

src/rawtoaces_util/transform_cache.h
antond-weta: that is expected (for now)

tests/testCache.cpp
antond-weta: fixed

soswow (approved):
wow. you've managed increase coverage by 1% 👏

soswow (approved):

PR #237, opened 2026-01-12
Refactor error handling to separate library and CLI error messages

aabhinavvvvvvv:
## Description

This PR refactors error handling in the `rawtoaces_util` library to separate library-level and CLI-level error messages, addressing the issue where library code was referencing CLI-specific flags in error messages. When using the `ImageConverter` library directly (not via the command-line tool), error messages were referencing CLI-specific flags that don't exist in the library context

## Tests

This is a refactoring that maintains backward compatibility. The existing test suite should cover the functionality. The changes primarily affect error message formatting and add a new status property, which doesn't change the API behavior, only enhances error reporting.

## Checklist:

- [ x] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/rawtoaces/blob/main/docs/CONTRIBUTING.md).
- [ x] I have updated the documentation, if applicable. (Check if there is no
  doesn't change the API.)
- [ x] I have ensured that the change is tested somewhere in the testsuite
- [ x] My code follows the prevailing code style of this project. If I haven't
  nonconforming.

### DIFF
--- a/tests/test_image_converter.cpp
+++ b/tests/test_image_converter.cpp
@@ -2357,12 +2357,10 @@ void test_last_error_message_colour_transforms_matrix_from_illuminant()
     bool has_relevant_error =
         converter.last_error_message.find( "Colour space transform" ) !=
             std::string::npos ||
-        converter.last_error_message.find( "matrix" ) !=
-            std::string::npos ||
+        converter.last_error_message.find( "matrix" ) != std::string::npos ||
         converter.last_error_message.find( "Failed to calculate" ) !=
             std::string::npos ||
-        converter.last_error_message.find( "spectral" ) !=
-            std::string::npos;
+        converter.last_error_message.find( "spectral" ) != std::string::npos;
     OIIO_CHECK_ASSERT( has_relevant_error );
 }
 


aabhinavvvvvvv:
> This is a good start, but it doesn't completely separate error output from the library. You still rely on stderr in the library to get any informative error messages, right? You could store the error message alongside the error code, but then the question would be - do we still need the error code at all?

You’re right that the current version still relies on stderr. I think we can keep ImageConverter::Status as a programmatic error category and add a last_error_message (or similar) field on ImageConverter that gets set whenever a bool‑returning method fails. Remove the CLI‑specific text (like --overwrite) from the library layer and instead have the CLI read status and last_error_message and print messages and library users can handle errors entirely via the API without parsing stderr, while the CLI still provides helpful message. If this sounds fine then i can update the PR accordingly

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1341,11 +1341,19 @@ bool ImageConverter::configure(
...
+        status = Status::ConfigurationError;
         return false;
     }
 
     fix_metadata( image_spec );
-    return configure( image_spec, options );
+    bool config_result = configure( image_spec, options );
+    if ( !config_result )
+    {
+        // status is already set by the other configure method
```
soswow: why wouldn't that other configure method set success status too?

Also, why do we need to set it explicitly down below? I thought  we set it by default. And so we only need to set it once - when error happens. In that case this whole set of changes is not required.

aabhinavvvvvvv: You're right..  I updated the string based configure() wrapper to only set status = ConfigurationError when ImageInput::open() fails, and otherwise just return configure(image_spec, options);

aabhinavvvvvvv:
> Could you please rebase your branch from main, some of the code you modified in the last commit is outdated. Also please add unit tests for the newly added functionality, including the python bindings. Thanks!

I've made the changes and added tests. Thanks

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1993,12 +2045,13 @@ bool ImageConverter::process_image( const std::string &input_filename )
     usage_timer.reset();
     if ( !save_image( output_filename, buffer ) )
     {
-        std::cerr << "Failed to save the file: " << output_filename
-                  << std::endl;
+        // status and _last_error_message are already set by save_image
```
soswow: do we need these comments?

aabhinavvvvvvv: You're right — those comments aren't needed. I added them during implementation for clarity, but the code is self-explanatory when a method returns false. I can remove them if you wish

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1813,20 +1865,20 @@ bool ImageConverter::make_output_path(
...
+        status = Status::Success;
+        _last_error_message.clear();
         return true;
     }
     catch ( const std::exception &e )
     {
-        std::cerr << "ERROR: Invalid path format '" << path << "': " << e.what()
-                  << std::endl;
+        status = Status::InvalidPath;
+        _last_error_message = std::string("Invalid path format '") + path + "': " + e.what();
```
soswow: Maybe this is c++ way, I don't know. But in my dayjob with JS (and JAva, I'd think) we would just have subclasses of `Error` which you initiate by providing specific description in constructor, hence capturing both type and details. like `new InvalidPath(std::string("Invalid path format '") + path + "': " + e.what())` in your case. It feel funky to have these two lines everywhere

tests/test_image_converter.cpp
```diff
@@ -2117,6 +2117,128 @@ void test_main_parse_parameters_failure()
...
+    
+    // First we process the file successfully to create output
+    bool first_result = converter.process_image( test_file );
+    if ( !first_result )
+    {
+        return;
+    }
+
+    // Clear any previous error message
+    converter.process_image( "dummy" );
```
antond-weta: What is the purpose of this?

src/rawtoaces_util/image_converter.cpp
aabhinavvvvvvv: I see the point about the repetition. In C++, using an enum for status codes and a separate string for messages is common, though it does require setting both. I can add a helper method like set_error(Status code, string& message) to reduce the duplication

soswow: yeah, I agree. I understand now. I think set_error will make it cleaner a bit.

tests/test_image_converter.cpp
aabhinavvvvvvv: That line was a mistake on my part. I initially thought it would clear any previous error state, but process_image("dummy") would set an error. Since the first process_image will succeeed and clear error message that line isn't needed. I'll remove it.

antond-weta: Would there ever be a need to clear the error message? If so, is there a functionality for that?

aabhinavvvvvvv: I think there could be legitimate use cases for explicitly clearing the error message, for example:
- Before any new set of operations where you want to check if any errors occurred during the last set
- When reusing an ImageConverter instance and wanting to reset its error state

While the default behavior should be that last_error_message persists until a new error occurs (representing "the last error encountered"), it makes sense to provide an explicit method for users who want to clear it.
something like 
```cpp
void clear_error() {
    status = Status::Success;
    _last_error_message.clear();
}
```

antond-weta:
Please address the issues raised in the comments above. 
Thanks.

include/rawtoaces/image_converter.h
```diff
@@ -363,6 +402,9 @@ class ImageConverter
     std::vector<std::vector<double>> _idt_matrix;
     std::vector<std::vector<double>> _cat_matrix;
     std::vector<double>              _wb_multipliers;
+    
+    // Error message from the most recent method call that returned false.
+    std::string _last_error_message;
```
antond-weta: I would just make this public and ditch the getter. If the error message property is readonly, you'd need to add a method to reset it. The status code property is public, so might have the message public as well.

src/rawtoaces_util/colour_transforms.cpp
```diff
@@ -12,30 +12,23 @@ namespace rta
 namespace util
 {
 
-void print_data_error( const std::string &data_type )
```
antond-weta, line 15: Why not keep this function and add the output message parameter instead? This would remove quite a bit of code duplication.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -273,6 +273,7 @@ CameraIdentifier get_camera_identifier(
     return { camera_make, camera_model };
 }
 
+// Removed print_data_error - errors are now communicated through ImageConverter::status and last_error_message
```
antond-weta: please remove this comment

antond-weta:
Please rebase your branch. It is not mergeable due to conflicts. Feel free to squash your commits if it's easier that way.
The unit tests are failing on CI.
The clang formatting check is failing.

antond-weta:
The code looks mostly fine now.

The unit tests are failing, because the existing tests rely on capturing the std::cerr to check for error messages. Since the error messages are not being written to std::cerr, the unit tests need to be rewritten to check the last_error_message instead.

aabhinavvvvvvv:
I've verified locally:
All 15 tests pass (100%)
All files pass clang-format
I also fixed the clang-format issue in tests/config_tests/util/test_installed_util.cpp (each array element now on its own line).

The remaining CI failures are infrastructure issues:
Build scripts don't have execute permission (permission denied errors)

antond-weta:
That's better, thanks!
Could you please apply the same trick to https://github.com/AcademySoftwareFoundation/rawtoaces/blob/730fa59a78e8edd6f67b8d7e765b9217c3abd2d4/.github/workflows/coverage.yml#L21

tests/test_image_converter.cpp
```diff
@@ -2093,6 +2128,246 @@ void test_main_parse_parameters_failure()
...
+    converter.settings.illuminant = "D65";
+    converter.settings.matrix_method =
+        ImageConverter::Settings::MatrixMethod::Spectral;
+    converter.settings.overwrite = true;
+
+    std::string test_file = std::filesystem::absolute( dng_test_file ).string();
+    bool        result    = converter.process_image( test_file );
+
+    // This might fail due to missing training data needed for IDT matrix calculation
+    if ( !result )
```
soswow: not a fan of conditional assertion in a test. do we really need it? All data required for test fixture should be present. if it doesn't - maybe it should fail.

soswow (approved):

PR #241, opened 2026-01-17
Increase colour_transform coverage

soswow:
more tests


src/rawtoaces_util/transform_cache.cpp
```diff
@@ -14,7 +14,7 @@ std::ostream &
 operator<<( std::ostream &os, const rta::core::Metadata &data [[maybe_unused]] )
 {
     return os << std::string( "<Metadata>" );
```
antond-weta, line 16:
```suggestion
    os << std::string( "<Metadata>" );
    return os;
```
would this help?

soswow: =( 

Image

PR #242, opened 2026-01-20
[feature] exiftool

antond-weta:
## Description

Adds runtime dependency on exiftool for fetching missing metadata. Exiftool gets executed as a subprocess to avoid any licensing issues. This is a pre-requisite for the lens functionality I'm working on. 

In the current state this code should never get executed, as without any lens corrections the only medatada this code could potentially fetch is camera make and model, which should be available from libraw for any valid file. 

## Tests

yes

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_util/exiftool.cpp
+++ b/src/rawtoaces_util/exiftool.cpp
@@ -139,7 +139,9 @@ bool fetch_metadata(
                              { "FNumber", { "aperture", true } },
                              { "FocalLength", { "focalLength", true } } };
 
-    std::string command = exiftool_path + " -S ";
+    // Adding the file name to the requested attributes, so the result of a
+    // successful call is never empty.
+    std::string command = exiftool_path + " -S -FileName";
     for ( auto key: keys )
     {
         if ( oiio_to_exiftool.count( key ) )
@@ -177,17 +179,22 @@ bool fetch_metadata(
             std::string exiftool_key = line.substr( 0, pos );
             std::string value        = line.substr( pos + 2 );
 
-            auto              &map_entry = exiftool_to_oiio.at( exiftool_key );
-            const std::string &oiio_key  = std::get<0>( map_entry );
-            bool               to_float  = std::get<1>( map_entry );
+            if ( exiftool_key != "FileName" )
+            {
+                assert( exiftool_to_oiio.count( exiftool_key ) );
 
-            if ( to_float )
-            {
-                spec[oiio_key] = std::stof( value );
-            }
-            else
-            {
-                spec[oiio_key] = value;
+                auto &map_entry = exiftool_to_oiio.at( exiftool_key );
+                const std::string &oiio_key = std::get<0>( map_entry );
+                bool               to_float = std::get<1>( map_entry );
+
+                if ( to_float )
+                {
+                    spec[oiio_key] = std::stof( value );
+                }
+                else
+                {
+                    spec[oiio_key] = value;
+                }
             }
         }
     }
--- a/tests/test_image_converter.cpp
+++ b/tests/test_image_converter.cpp
@@ -2132,6 +2132,13 @@ void test_main_no_files_processed()
 
 void test_fetch_missing_metadata()
 {
+    std::cout << std::endl << "test_fetch_missing_metadata()" << std::endl;
+
+#if defined( WIN32 ) || defined( WIN64 )
+    const char *exiftool_path = "..\\..\\exiftool\\exiftool.exe";
+    set_env_var( "RAWTOACES_EXIFTOOL_PATH", exiftool_path );
+#endif
+
     rta::util::ImageConverter converter;
     OIIO::ImageSpec           spec;
 
@@ -2150,6 +2157,14 @@ void test_fetch_missing_metadata()
     OIIO_CHECK_EQUAL(
         spec.get_string_attribute( "cameraModel" ), "NIKON D200" );
 
+    spec.extra_attribs.remove( "cameraMake" );
+    spec.extra_attribs.remove( "cameraModel" );
+    // The dng doesn't contain these attributes, but the call should not fail.
+    result = fetch_missing_metadata( dng_test_file, converter.settings, spec );
+    OIIO_CHECK_ASSERT( result );
+
+    spec.extra_attribs.remove( "cameraMake" );
+    spec.extra_attribs.remove( "cameraModel" );
     result =
         fetch_missing_metadata( "wrong_filename", converter.settings, spec );
     OIIO_CHECK_ASSERT( !result );


lgritz:
No, no, I didn't mean *instead* of what you have here.

This PR just reminded me of that incomplete work and I was curious if you think there's any use to my picking it back up again.

build_scripts/install_deps_windows.bash
```diff
@@ -19,3 +19,7 @@ vcpkg install \
 VCPKG_PYTHON="C:/vcpkg/installed/x64-windows/tools/python3/python.exe"
 "$VCPKG_PYTHON" -m ensurepip --upgrade
 "$VCPKG_PYTHON" -m pip install pytest
+
+curl --silent --output ./exiftool.zip https://exiftool.org/exiftool-13.45_64.zip
+unzip ./exiftool.zip
+mv ./exiftool-13.45_64/exiftool\(-k\).exe ./exiftool-13.45_64/exiftool.exe
```
soswow: are we intentionally keeping version as part of moving? We could move it into `/exiftool/exiftool.exe` and save number of lines needed to be change when we need to update the version. But I Can see how you could made it intentional.

src/rawtoaces_util/exiftool.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    return "";
+}
+
+bool execute( const std::string &command, std::stringstream &stream )
+{
+    constexpr size_t buf_size = 255;
+    char             buffer1[buf_size];
+    //    std::stringstream stream;
```
soswow: comment line can be killed?

src/rawtoaces_util/exiftool.cpp
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    return "";
+}
+
+bool execute( const std::string &command, std::stringstream &stream )
+{
+    constexpr size_t buf_size = 255;
+    char             buffer1[buf_size];
```
soswow: any reason it has 1 on the end? maybe `stream_buffer` or somthing?

src/rawtoaces_util/exiftool.cpp
```diff
@@ -0,0 +1,200 @@
...
+    const std::vector<std::string> &keys )
+{
+
+    std::string exiftool_path = find_exiftool();
+
+    if ( exiftool_path.empty() )
+    {
+        std::cerr
+            << "Exiftool not found, please provide the path to "
+            << "exiftool binary via the RAWTOACES_EXIFTOOL_PATH environment "
```
soswow: We don't want to advertise the PATH trick? It can be part of it as well, right?

src/rawtoaces_util/exiftool.cpp
```diff
@@ -0,0 +1,200 @@
...
+        else
+        {
+            std::cerr << "Exiftool: unknown key " << key << std::endl;
+            return false;
+        }
+    }
+
+    command += " " + path;
+
+    std::stringstream stream2;
```
soswow: curious about number of the end of varibles?

src/rawtoaces_util/exiftool.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if ( !execute( command, stream2 ) )
+    {
+        std::cerr
+            << "Failed to execute exiftool. Please provide the path to "
+            << "exiftool binary via the RAWTOACES_EXIFTOOL_PATH environment "
```
soswow: 1) Would prob. be useful to see error message that exiftool returns
2) Didn't we already have a place where we catch situation where executable is not found?

src/rawtoaces_util/exiftool.cpp
```diff
@@ -0,0 +1,200 @@
...
+    std::string                        line;
+    while ( std::getline( stream2, line ) )
+    {
+        auto pos = line.find( ": " );
+        if ( pos != line.npos )
+        {
+            std::string exiftool_key = line.substr( 0, pos );
+            std::string value        = line.substr( pos + 2 );
+
+            auto              &map      = exiftool_to_oiio.at( exiftool_key );
```
soswow: variable name might be confusing for a reader since `map` usually reserved for object that have set of key/value pairs. Where is this is result of mapping. So, it's `result_pair` maybe? or `oiio_key_and_to_float_pair`

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1258,6 +1264,38 @@ void fix_metadata( OIIO::ImageSpec &spec )
     }
 }
 
+bool fetch_missing_metadata(
+    const std::string              &input_path,
+    const ImageConverter::Settings &settings,
+    OIIO::ImageSpec                &spec )
+{
+    if ( settings.disable_exiftool )
+        return true;
```
soswow, line 1273: is that semantically / logically correct? If goal of this function is to fetch missing metadata - we exiting here without achieving it. So, should it be `true` ?

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1258,6 +1264,38 @@ void fix_metadata( OIIO::ImageSpec &spec )
...
+
+    std::vector<std::string> keys_to_check;
+
+    if ( settings.custom_camera_make.empty() )
+        keys_to_check.push_back( "cameraMake" );
+    if ( settings.custom_camera_model.empty() )
+        keys_to_check.push_back( "cameraModel" );
+
+    std::vector<std::string> keys_to_fetch;
```
soswow, line 1282: we already did that on 1275.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1277,6 +1315,12 @@ bool ImageConverter::configure(
     }
 
     fix_metadata( image_spec );
+    result = fetch_missing_metadata( input_filename, settings, image_spec );
+    if ( !result )
+    {
+        return false;
```
soswow, line 1321: do we want to have a warning/error explaining why we failing ::configure?

tests/test_utils.cpp
soswow, line 1: ❤️

tests/testExiftool.cpp
```diff
@@ -0,0 +1,118 @@
...
+#elif defined( __APPLE__ )
+    const char *exiftool_path = "/opt/homebrew/bin/exiftool";
+#else
+    const char *exiftool_path = "/usr/bin/exiftool";
+#endif
+    set_env_var( "RAWTOACES_EXIFTOOL_PATH", exiftool_path );
+    unset_env_var( "PATH" );
+
+    std::string output = check( false, { "bad_key" } );
+    ASSERT_CONTAINS( output, "Exiftool: unknown key " );
```
soswow, line 106: should there "bad_key" mentioned in expected text?

build_scripts/install_deps_windows.bash
antond-weta: fixed

src/rawtoaces_util/exiftool.cpp
antond-weta: fixed

src/rawtoaces_util/exiftool.cpp
antond-weta: fixed

src/rawtoaces_util/exiftool.cpp
antond-weta: fixed

src/rawtoaces_util/exiftool.cpp
antond-weta: fixed

src/rawtoaces_util/exiftool.cpp
antond-weta: There is no error message at this stage. The output from exiftool gets processed below that line.
The binary not found and failed to execute are not the same situation.

src/rawtoaces_util/exiftool.cpp
antond-weta: fixed

src/rawtoaces_util/image_converter.cpp
antond-weta: I'm happy with that. `true` means no error. Having 3-state output is too much work for a local helper function.

src/rawtoaces_util/image_converter.cpp
antond-weta: not really?

src/rawtoaces_util/image_converter.cpp
antond-weta: That call should never fail there for any valid file. If it does, we're having much bigger problems, and will be seeing other errors.

The same function will be used in another place, where it can fail, there will be proper error reporting there.

src/rawtoaces_util/image_converter.cpp
soswow: indeed on closer inspection.

soswow (approved):

PR #247, opened 2026-02-09
refactor logging

antond-weta:
## Description

Cleaned up more `std::cerr` output statements. I believe all error messages are fixed now. The verbose messages are still being written to `std::cerr`. 

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_core/rawtoaces_core.cpp
+++ b/src/rawtoaces_core/rawtoaces_core.cpp
@@ -386,7 +386,7 @@ bool SpectralSolver::find_illuminant( const vector<double> &wb )
             SpectralData     &illuminant_data = _all_illuminants.emplace_back();
             const std::string type = "d" + std::to_string( cct / 100 );
 
-            bool success = generate_illuminant(
+            [[maybe_unused]] bool success = generate_illuminant(
                 cct, type, true, illuminant_data, last_error_message );
             assert( success );
         }
@@ -397,7 +397,7 @@ bool SpectralSolver::find_illuminant( const vector<double> &wb )
             SpectralData     &illuminant_data = _all_illuminants.emplace_back();
             const std::string type            = std::to_string( cct ) + "k";
 
-            bool success = generate_illuminant(
+            [[maybe_unused]] bool success = generate_illuminant(
                 cct, type, false, illuminant_data, last_error_message );
             assert( success );
         }


src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -372,15 +385,17 @@ bool SpectralSolver::find_illuminant( const vector<double> &wb )
         {
             SpectralData     &illuminant_data = _all_illuminants.emplace_back();
             const std::string type = "d" + std::to_string( cct / 100 );
-            generate_illuminant( cct, type, true, illuminant_data );
+            generate_illuminant(
```
soswow: we are not using result of `generate_illuminant` here?

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -129,12 +120,15 @@ void calculate_blackbody_SPD( const int &cct, Spectrum &spectrum )
 /// @param type Type of light source (e.g. "d50", "d65", "d75", "A", "B", "C", "D50", "D65", "D75")
 /// @param is_daylight True if the light source is a daylight source, false if it is a blackbody source
 /// @param illuminant Reference to SpectralData object to fill with generated illuminant data
+/// @param error_message a destination for any potential error message.
+/// @result `true` if generated successfully.
 /// @pre cct is in valid range for the specified illuminant type
-void generate_illuminant(
+bool generate_illuminant(
```
soswow, line 126: It's probably one of those c++ patterns, so take it more of a question. 
Alternative to this would be function that returns an error string. So, two in one - if string is not empty === unsuccessful result etc.
I can see that this pattern would need to get following later on as well, since other places that use this function return boolean as a result still.
anyway, just thoughts out loud.

src/rawtoaces_util/colour_transforms.cpp
```diff
@@ -204,7 +205,8 @@ bool solve_multipliers_from_illuminant(
 
     if ( !solver.calculate_WB() )
     {
-        error_message = "Failed to calculate white balance multipliers";
+        error_message = "Failed to calculate white balance multipliers. " +
+                        solver.last_error_message;
```
soswow, line 209: /nit Do we care that there is a change error_message will end with a space? I am sure it is not critical.

src/rawtoaces_util/exiftool.cpp
```diff
@@ -105,17 +105,18 @@ bool execute( const std::string &command, std::stringstream &stream )
...
-        std::cerr
-            << "Exiftool not found, please make sure that its location is "
-            << "available in PATH. Alternatively you can provide the path to "
-            << "the exiftool binary via the RAWTOACES_EXIFTOOL_PATH "
-            << "environment variable." << std::endl;
+        error_message =
+            "Exiftool not found, please make sure that its location is "
+            "available in PATH. Alternatively you can provide the path to the "
+            "exiftool binary via the RAWTOACES_EXIFTOOL_PATH environment "
+            "variable.";
```
soswow, line 119: /nit observation - it seems like we are using both methods of string concatenation. Should we do just one for consistency?

src/rawtoaces_core/rawtoaces_core.cpp
antond-weta: added assertion

src/rawtoaces_core/rawtoaces_core.cpp
antond-weta: We only use this in private functions. This way we can always reconsider and change later if there is a better way.
There is one exception where I added the error_message param in the public method of the core library, but that is a lesser evil, as an optional parameter doesn't break API compatibility.

src/rawtoaces_util/colour_transforms.cpp
antond-weta: Avoiding an extra space in a log message doesn't worth writing code to me.

src/rawtoaces_util/exiftool.cpp
antond-weta: yes we should. too minor to make an effort.

PR #252, opened 2026-02-16
[logging] refactoring; additional checks for the database dirs

antond-weta:
## Description

- Add additional check on the database directories: if none of them contain "camera", "cmf", "observer", or "training" subfolder, an error message is shown.
- Remove the check for the requested illuminant being present in the DB from `ImageConverter::parse_parameters`. That method is only being called when executed from command line. Also a user could potentially re-configure the DB paths after calling `ImageConverter::parse_parameters`. Instead a check for the missing illuminant is added in the image processing loop, if such error is caught rawtoaces errors out instead of processing the next image.
- The command line parameter hints which were removed from the library level in the previous refactor are now re-introduced on the app level.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/tests/test_image_converter.cpp
+++ b/tests/test_image_converter.cpp
@@ -541,6 +541,21 @@ void test_convert_linux_path_to_windows_path()
     OIIO_CHECK_EQUAL( result, "c:\\path1;c:\\path2;c:\\path3" );
 }
 
+/// Tests empty path passed to make_output_path
+void test_empty_input_path()
+{
+    std::cout << "\n" << __FUNCTION__ << std::endl;
+
+    std::string    path = "";
+    ImageConverter converter;
+    bool           success = converter.make_output_path( path );
+    OIIO_CHECK_ASSERT( !success );
+    OIIO_CHECK_ASSERT(
+        converter.status == ImageConverter::Status::EmptyInputFilename );
+    ASSERT_CONTAINS(
+        converter.last_error_message, "Empty input path provided." );
+}
+
 /// Tests fix_metadata with both Make and Model attributes
 void test_fix_metadata_both_attributes()
 {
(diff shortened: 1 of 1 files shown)


src/rawtoaces/main.cpp
```diff
@@ -66,8 +94,22 @@ int main( int argc, const char *argv[] )
...
+                    std::string error_message = converter.last_error_message;
+                    update_error_message( error_message );
+                    std::cerr << "  Reason: " << error_message << std::endl;
+
+                    if ( converter.status ==
+                         rta::util::ImageConverter::Status::ConfigurationError )
+                    {
+                        if ( error_message.find(
+                                 "Failed to find illuminant type" ) !=
+                             std::string::npos )
```
soswow, line 106: /nit would joining these two if make code less comples just a bit?

src/rawtoaces_util/image_converter.cpp
```diff
@@ -233,6 +233,50 @@ std::vector<std::string> database_paths( const std::string &override_path = "" )
...
+    if ( !found_database )
+    {
+        error_message =
+            "Spectral measurements database not found. "
+            "Please provide the database path using the "
+            "RAWTOACES_DATA_PATH environment variable.";
+        return false;
+    }
+
+    return true;
```
soswow, line 275: just to confirm the I understand the logic of this function correctly and that is intended behaviour - it goes through the list of directories and inside each it is looking for any of "camera", "cmf", "illuminant", "training". If it finds any of the given directories contains at least one of these from the list - it returns true. Corrent? Reason I am asking is I would assume you want to find a directory that has all four?

antond-weta: Yes, that is the intended (by me) behaviour. We don't need each of the folders to contain all 4 data types. The user may want to provide a supplementary database of just camera measurements. We could change the logic to error out only in the case if any of the four hasn't been found once, but that's unnecessary in my opinion for a few reasons:
- We expect the official `rawtoaces-data` DB to be present in the path, and it always comes with all 4 data types.
- Not all modes require all files. If you only need white balancing to an illuminant, only the camera files are required.
- This is just an additional check to make error reporting a bit more user-friendly. If a required file can't be found in the database, you still get the corresponding error down the line.

PR #255, opened 2026-03-04
[feat] lens correction

antond-weta:
## Description

Implements lens correction functionality

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/bindings/py_util.cpp
+++ b/src/bindings/py_util.cpp
@@ -275,7 +275,7 @@ void util_bindings( nanobind::module_ &m )
             "ConfigurationError", ImageConverter::Status::ConfigurationError )
         .value( "ReadError", ImageConverter::Status::ReadError )
         .value(
-            "LensCorrectionError", ImageConverter::Status::LensCoorectionError )
+            "LensCorrectionError", ImageConverter::Status::LensCorrectionError )
         .value(
             "MatrixApplicationError",
             ImageConverter::Status::MatrixApplicationError )


src/rawtoaces_util/exiftool.cpp
```diff
@@ -136,7 +136,6 @@ bool fetch_metadata(
         exiftool_to_oiio = { { "Make", { "cameraMake", false } },
                              { "Model", { "cameraModel", false } },
                              { "LensID", { "lensModel", false } },
-                             { "LensModel", { "lensModel", false } },
```
soswow, line 139: Curious why we removing it?

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1356,6 +1517,195 @@ bool fetch_missing_metadata(
...
+        {
+            return false;
+        }
+
+        std::string lens_make = settings.custom_lens_make;
+        // Do not fail if the lens make is missing. We can work around that.
+        fetch_lens_parameter(
+            lens_make,
+            src_spec,
+            "lenMake",
```
soswow: len**s**Make

src/rawtoaces_util/lens_correction.cpp
```diff
@@ -0,0 +1,632 @@
...
+        error_message = "Lens correction data for the camera make '" +
+                        camera_make + "' and model '" + camera_model +
+                        "' not found " + "in the database.";
+        return nullptr;
+    }
+
+    auto cam = cameras[0];
+
+    const lfLens **lenses =
+        Database->FindLenses( cam, NULL, lens_model.c_str() );
```
soswow: should `lens_make` be used here instead of NULL?

```
const char* lens_make_ptr =
    lens_make.empty() ? nullptr : lens_make.c_str();

const lfLens** lenses =
    Database->FindLenses(cam, lens_make_ptr, lens_model.c_str());
```

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1356,6 +1517,195 @@ bool fetch_missing_metadata(
...
+                return false;
+            }
+        }
+
+        if ( settings.lens_correction_types &&
+             ImageConverter::Settings::LensCorrectionType::Aberration )
+        {
+            if ( !apply_aberration_map(
+                     dst,
+                     src,
```
soswow: at each state of these three calls we give `dst` and `src` where is `src` is seem to be the same object that came in originally. The only reason it works if because our internal called uses same object for dst and src. But this is public API and hence we should handle the case where these two are not the same.

```
if ( &dst != &src )
   dst.copy( src );
...
            if ( !apply_vignette_map(
                     dst,
                     dst,
...
            if ( !apply_aberration_map(
                     dst,
                     dst,
...
            if ( !apply_distortion_map(
                     dst,
                     dst,
```
or some more robust way.

Also, it seems like you have tests `void test_apply_lens_correction()` where those two are not the same, so am I correct assuming the result would be only distortion map applied?

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1356,6 +1517,195 @@ bool fetch_missing_metadata(
...
+            focus_distance = settings.custom_focus_distance;
+            fetch_lens_parameter(
+                focus_distance,
+                src_spec,
+                "focus",
+                "focus distance value",
+                last_error_message );
+        }
+
+        OIIO::ParamValueList options;
```
soswow: is it used?

src/bindings/py_util.cpp
soswow, line 288: we need to add 
```
.value(
            "DatabaseNotFound", ImageConverter::Status::DatabaseNotFound )
```

src/docs/lens_correction.rst
```diff
@@ -0,0 +1,131 @@
...
+Request all supported types of lens correction, make the correction mandatory,
+override all camera/lens info used for correction:
+
+.. tabs::
+  .. tab:: Shell
+    .. code-block:: bash
+
+      rawtoaces                                    \
+      --lens-correction a                          \
+      --require-lens correction                    \
```
soswow:
```suggestion
      --require-lens-correction                    \
```

src/docs/lens_correction.rst
```diff
@@ -0,0 +1,131 @@
...
+      --custom-camera-make "Canon"                 \
+      --custom-camera-model "EOS R5"               \
+      --custom-lens-make "Canon"                   \
+      --custom-lens-model "RF 24-105mm F4L IS USM" \
+      --custom-aperture 4.0                        \
+      --custom-focal-length 35.0                   \
+      --custom-focus_distance 240.0                \
```
soswow:
```suggestion
      --custom-focus-distance 240.0                \
```

tests/python/test_image_converter.py
```diff
@@ -377,7 +377,30 @@ def test_settings_attributes(self):
...
+        assert (converter.settings.lens_correction_types & rawtoaces.ImageConverter.Settings.LensCorrectionType.Aberration).value == 0
         
+        converter.settings.require_lens_correction = True
+        assert converter.settings.require_lens_correction == True
+    
+        converter.settings.custom_lens_make = "custom_lens_make"
+        assert converter.settings.custom_lens_make == "custom_lens_make"
+    
+        converter.settings.custom_lens_make = "custom_lens_model"
+        assert converter.settings.custom_lens_make == "custom_lens_model"
```
soswow:
```suggestion
        converter.settings.custom_lens_model = "custom_lens_model"
        assert converter.settings.custom_lens_model == "custom_lens_model"
```

src/rawtoaces_util/exiftool.cpp
antond-weta: Exiftool seems to be inconsistent with this field for some cameras. This makes it difficult to decide what to do when both LensID and LensModel are present and hold different value.  

BTW I'm expecting this code to be tweaked a fair bit as this functionality gets more use. Collecting metadata from raw images is always a balancing act.

antond-weta: I've refactored that further to make it a bit more robust. The lensModel thingy is back!

src/rawtoaces_util/image_converter.cpp
antond-weta: fixed

src/rawtoaces_util/lens_correction.cpp
antond-weta: fixed

src/rawtoaces_util/image_converter.cpp
antond-weta: good catch! fixed.

src/rawtoaces_util/image_converter.cpp
antond-weta: fixed

src/bindings/py_util.cpp
antond-weta: fixed

src/docs/lens_correction.rst
antond-weta: fixed

src/docs/lens_correction.rst
antond-weta: fixed

tests/python/test_image_converter.py
antond-weta: fixed

src/rawtoaces_util/image_converter.cpp
soswow: not sure. I think I understand what LLM is saying, but I rather paste it as-is so nothing is lost in translation:

`[image_converter.h (line 374)]` documents apply_lens_correction(dst, src) as supporting distinct buffers, but the implementation starts with `src_ptr = &src` and only switches to dst after a prior stage succeeds `[image_converter.cpp (line 1636)]`. If the first enabled stage is distortion or aberration, the lower-level code depends on dst being already initialized: distortion takes its temp spec from `dst_buffer.spec()` `[lens_correction.cpp (line 461)]`, and aberration actually warps from dst_buffer instead of src_buffer `[lens_correction.cpp (line 600)]`, `[lens_correction.cpp (line 612)]`. Existing success-path tests are in-place for distortion/aberration `[testLensCorrection.cpp (line 261)]`, `[testLensCorrection.cpp (line 317)]`, while the only out-of-place success test is vignetting `[testLensCorrection.cpp (line 466)]`.

Suggested fix: initialize a working buffer from src when `dst != src`, chain every enabled stage against the previous output, and add explicit out-of-place success tests for distortion-only, aberration-only, and combined correction.

src/rawtoaces_util/image_converter.cpp
```diff
@@ -1338,6 +1478,27 @@ bool fetch_missing_metadata(
     if ( settings.custom_camera_model.empty() )
         keys_to_check.push_back( "cameraModel" );
 
+#if ( RTA_ENABLE_LENSFUN )
+    if ( settings.lens_correction_types !=
+         ImageConverter::Settings::LensCorrectionType::None )
+    {
+        if ( settings.custom_lens_model.empty() )
+            keys_to_check.push_back( "lensModel" );
```
soswow, line 1486: shuould we be doing same for `lensMake`?

src/rawtoaces_util/exiftool.cpp
```diff
@@ -102,61 +102,55 @@ bool execute( const std::string &command, std::stringstream &stream )
...
+
     // Mapping from OIIO attribute names to ExifTool attribute names.
     const std::map<std::string, std::string> oiio_to_exiftool = {
-        { "cameraMake", "Make" },
-        { "cameraModel", "Model" },
-        { "lensModel", "LensID" },
-        { "aperture", "FNumber" },
-        { "focalLength", "FocalLength" }
+        { "cameraMake", "Make" },     { "cameraModel", "Model" },
+        { "lensModel", "LensModel" }, { "lensID", "LensID" },
```
soswow, line 127: do we need mapping for lensMake?

src/rawtoaces/main.cpp
```diff
@@ -15,6 +13,15 @@ void update_error_message( std::string &error_message )
         { "Missing the camera model name in the file metadata.",
           " You can provide a custom value using the --custom-camera-model "
           "parameter." },
+        { "Missing the lens model name in the file metadata.",
```
soswow, line 16: actual message is `"Missing the <property> info in the file metadata."`

src/bindings/py_util.cpp
```diff
@@ -228,8 +246,22 @@ void util_bindings( nanobind::module_ &m )
...
+    nanobind::enum_<ImageConverter::Settings::LensCorrectionType>(
+        settings, "LensCorrectionType", nanobind::is_flag() )
+        .value(
+            "Aberration",
+            ImageConverter::Settings::LensCorrectionType::Aberration )
+        .value(
+            "Distortion",
+            ImageConverter::Settings::LensCorrectionType::Distortion )
+        .value(
+            "Vignetting",
```
soswow, line 258: There is also `None`
```
enum class LensCorrectionType
        {
            None       = 0,
            Aberration = 1 << 0,
            Distortion = 1 << 1,
            Vignetting = 1 << 2
        };
```

src/rawtoaces_util/image_converter.cpp
antond-weta: exiftool does not report lens make as a separate field.

src/rawtoaces_util/exiftool.cpp
antond-weta: exiftool does not report lens make as a separate field.

src/bindings/py_util.cpp
antond-weta: It is not really needed, as you can easily assign zero. Enum class can not be implicitly cast to/from int, so we have the explicit zero option there. I was following the nanobind documentation for flag enum, they don't expose zero. 

I'm not a python dev though, we can add `None` later if it's required.

src/rawtoaces_util/image_converter.cpp
antond-weta: I've added unit tests for all permutations of correction types.
I've also changed the unit tests to have separate dst and src buffers. Since the apply_lens_correction() uses dst == src internally, this makes both options covered.

src/rawtoaces/main.cpp
antond-weta: fixed

src/bindings/py_util.cpp
soswow: setting zero wont work, but there is a workaround:
```python
def test_lens_correction_types_accept_zero(self):
        """Test that lens_correction_types can be cleared by assigning zero"""
        converter = rawtoaces.ImageConverter()

        converter.settings.lens_correction_types = (
            rawtoaces.ImageConverter.Settings.LensCorrectionType.Distortion |
            rawtoaces.ImageConverter.Settings.LensCorrectionType.Vignetting
        )
        # WON'T work
        # converter.settings.lens_correction_types = 0
        # 

        # Works
        converter.settings.lens_correction_types = ( 
            rawtoaces.ImageConverter.Settings.LensCorrectionType(0)
        )

        assert converter.settings.lens_correction_types.value == 0

PR #266, opened 2026-05-04
[analysis] fix static analysis errors

antond-weta:
## Checklist:

  doesn't change the API.)
  nonconforming.


src/rawtoaces_core/mathOps.h
```diff
@@ -229,17 +229,15 @@ std::vector<T> mulVector( std::vector<T> vct1, std::vector<T> vct2, int k = 3 )
     assert( k * cols == vct2.size() );
 
     std::vector<T> vct3( rows * cols );
-    T             *pA = &vct1[0];
-    T             *pB = &vct2[0];
-    T             *pC = &vct3[0];
```
soswow, line 234: not sure what's happening here, but I guess we don't need this initialisation

soswow (approved):

PR #267, opened 2026-05-04
[refactor] DNG solver

antond-weta:
## Description

Adds additional error checking to the DNG solver internals. 
Note: the errors mostly get ignored on the top level, as the public API `rta::core::DNGSolver::calculate_CAT_matrix` and `rta::core::DNGSolver::calculate_IDT_matrix` currently don't provide any means for error reporting apart from returning an empty matrix. I'm planning to deprecate those in favour of a safer option in one of the following PRs.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_core/mathOps.h
+++ b/src/rawtoaces_core/mathOps.h
@@ -77,6 +77,23 @@ T cross2d_scalar( const std::vector<T> &vectorA, const std::vector<T> &vectorB )
     return vectorA[0] * vectorB[1] - vectorA[1] * vectorB[0];
 }
 
+/// Calculate the determinant of a square matrix.
+///
+/// @param vMtx a matrix to calculate a determinant for
+/// @return the deternimant
+template <typename T> T determinant( const std::vector<std::vector<T>> &vMtx )
+{
+    assert( isSquare( vMtx ) );
+
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> m;
+    m.resize( vMtx.size(), vMtx[0].size() );
+    for ( Eigen::Index i = 0; i < m.rows(); i++ )
+        for ( Eigen::Index j = 0; j < m.cols(); j++ )
+            m( i, j ) = vMtx[i][j];
+
+    return m.determinant();
+}
+
 template <typename T>
 std::vector<std::vector<T>> invertVM( const std::vector<std::vector<T>> &vMtx )
 {
--- a/src/rawtoaces_core/rawtoaces_core.cpp
+++ b/src/rawtoaces_core/rawtoaces_core.cpp
@@ -1103,27 +1103,104 @@ double XYZ_to_color_temperature( const std::vector<double> &XYZ )
 /// @param matrix_end Second camera transformation matrix
 /// @return Interpolated camera transformation matrix
 /// @pre mired_start != mired_end to avoid division by zero
-std::vector<double> XYZ_to_camera_weighted_matrix(
+std::vector<std::vector<double>> XYZ_to_camera_weighted_matrix(
     const double              &mired_target,
     const double              &mired_start,
     const double              &mired_end,
     const std::vector<double> &matrix_start,
     const std::vector<double> &matrix_end )
 {
+    assert( matrix_start.size() == matrix_end.size() );
+    assert( matrix_start.size() % 3 == 0 );
+    assert( std::fabs( mired_start - mired_end ) > 1e-9 );
 
     double weight = std::max(
         0.0,
         std::min(
             1.0,
             ( mired_start - mired_target ) / ( mired_start - mired_end ) ) );
-    std::vector<double> result = subVectors( matrix_end, matrix_start );
-    scaleVector( result, weight );
-    result = addVectors( result, matrix_start );
+
+    size_t                           rows = matrix_start.size() / 3;
+    std::vector<std::vector<double>> result( rows );
+
+    for ( size_t row = 0, index = 0; row < rows; row++ )
+    {
+        result[row].resize( 3 );
+        for ( size_t col = 0; col < 3; col++ )
+        {
+            double val1      = matrix_start[index] * ( 1.0 - weight );
+            double val2      = matrix_end[index] * weight;
+            result[row][col] = val1 + val2;
+            index++;
+        }
+    }
 
     return result;
 }
 
-/// Find the optimal XYZ to camera transformation matrix using iterative optimization.
+/// Calculate the camera colour space to XYZ transform matrix.
+/// This function calls the XYZ_to_camera_weighted_matrix() and
+/// calculates the inverse matrix of its result.
+///
+/// @param mired_target Target Mired value for interpolation
+/// @param mired_start First reference Mired value (start of interpolation range)
+/// @param mired_end Second reference Mired value (end of interpolation range)
+/// @param matrix_start First camera transformation matrix
+/// @param matrix_end Second camera transformation matrix
+/// @param out_matrix Output parameter to hold the calculated matrix.
+/// @return True on success
+/// @pre mired_start != mired_end to avoid division by zero.
+/// @pre calculated XYZ to camera matrix must be inversible.
+bool camera_to_XYZ_weighted_matrix(
+    const double                     &mired_target,
+    const double                     &mired_start,
+    const double                     &mired_end,
+    const std::vector<double>        &matrix_start,
+    const std::vector<double>        &matrix_end,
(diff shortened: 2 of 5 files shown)


src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1357,21 +1475,27 @@ void get_camera_XYZ_matrix_and_white_point(
...
+
+    if ( get_camera_XYZ_matrix_and_white_point(
+             _metadata, camera_to_XYZ_matrix, camera_XYZ_white_point ) )
+    {
+        std::vector<double> output_RGB_to_XYZ_matrix =
+            matrix_RGB_to_XYZ( chromaticitiesACES );
+        std::vector<double> output_XYZ_white_point =
+            mulVector( output_RGB_to_XYZ_matrix, deviceWhiteV );
+        CAT_matrix =
+            calculate_CAT( camera_XYZ_white_point, output_XYZ_white_point );
```
remia, line 1495: Hi @antond-weta,

This CAT is computed from the camera white to ACES ~D60 white point, but below it gets multiplied by the XYZ D65 to ACES2065-1 matrix (including D65 to ~D60 CAT), should this be changed to target D65 or the below code changed to use `XYZ_acesrgb_3` instead of `XYZ_D65_acesrgb`?

Additionally, there may be an issue in the way the CAT is applied, as LibRaw seem to return a white balanced image with neutral encoded as D65 0.3127 0.3290 chromaticities. If we want to apply a CAT from camera white point to ACES ~D60, we would need to invert the white balancing applied in LibRaw (which is possible using rgb_cam, cam_mul, pre_mul already). On some edge case DNGs, the current code can results in pretty big color cast. Overall, seeing that LibRaw don't follow the DNG recommendation (it only uses 1 illuminant for the color processing, typically ColorMatrix2 from D65), I wonder if RawToACES should not try to compute a full IDT from CameraRGB instead of only the CAT part. But I can detail what I think in more detail in an issue if more appropriate.

antond-weta: Thanks Rémi, there should be `XYZ_acesrgb_3` instead of `XYZ_D65_acesrgb` indeed. I'll fix that as a separate PR as it is unrelated to this one.

Currently rawtoaces requests the images from libraw in XYZ when in DNG mode. I'd like to change that to be in the camera space, so we could have full control over the white balancing and input matrix steps. This should be a simple change, given that we calculate the weights and camera to XYZ matrix already.

antond-weta: I've tracked this down to https://github.com/AcademySoftwareFoundation/rawtoaces/pull/108
@remia, could you confirm if that is correct?

remia: Thanks @antond-weta yes your link seem to be part of the story, though the issue was probably even before that change as if I understand correctly this commit just changed the variable name but not the behaviour of the DNG conversion.

I agree with you on requesting raw output from LibRaw for DNGs, we might be missing the ForwardMatrix tags but otherwise can already have it working using ColorMatrix if I understand correctly. To be fair, even for RAWs I may want to use Adobe profiles and do something similar, but it is already possible with the custom matrix method so that's mostly fine (all this being when spectral data is unavailable obviously)

src/rawtoaces_core/mathOps.h
```diff
@@ -77,6 +77,23 @@ T cross2d_scalar( const std::vector<T> &vectorA, const std::vector<T> &vectorB )
     return vectorA[0] * vectorB[1] - vectorA[1] * vectorB[0];
 }
 
+/// Calculate the determinant of a square matrix.
+///
+/// @param vMtx a matrix to calculate a determinant for
+/// @return the deternimant
```
soswow:
```suggestion
/// @return the determinant
```

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1330,18 +1443,23 @@ std::vector<double> matrix_RGB_to_XYZ( const double chromaticities[][2] )
 /// @param metadata Camera metadata containing calibration and exposure information
 /// @param out_camera_to_XYZ_matrix Output camera to XYZ transformation matrix
 /// @param out_camera_XYZ_white_point Output camera white point in XYZ space
+/// @result true on success.
```
soswow, line 1444:
```suggestion
/// @return true on success.
```

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1105,27 +1105,104 @@ double XYZ_to_color_temperature( const std::vector<double> &XYZ )
...
+    out_matrix = invertVM( XYZ_to_camera_matrix );
+    return true;
+}
+
+/// Stack the rows of a matrix to convert from one-dimensional to
+/// two-dimensional representation.
+///
+/// @param matrix the input matrix as a one-dimensional array.
+/// @param columns the number of columns in the output matrix.
+/// @result the output matrix as a two-dimensional array.
```
soswow:
```suggestion
/// @return the output matrix as a two-dimensional array.
```

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1105,27 +1105,104 @@ double XYZ_to_color_temperature( const std::vector<double> &XYZ )
...
+/// @param mired_target Target Mired value for interpolation
+/// @param mired_start First reference Mired value (start of interpolation range)
+/// @param mired_end Second reference Mired value (end of interpolation range)
+/// @param matrix_start First camera transformation matrix
+/// @param matrix_end Second camera transformation matrix
+/// @param out_matrix Output parameter to hold the calculated matrix.
+/// @return True on success
+/// @pre mired_start != mired_end to avoid division by zero.
+/// @pre calculated XYZ to camera matrix must be inversible.
+bool camera_to_XYZ_weighted_matrix(
```
soswow, line 1154: suggestion from llm:
> It is a free function in `rta::core` with external linkage but only used inside `rawtoaces_core.cpp`. Consider static or an anonymous namespace so it does not become part of the public symbol surface of the translation unit.

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1137,87 +1214,123 @@ std::vector<double> XYZ_to_camera_weighted_matrix(
...
-    double mired_step = std::max( 5.0, ( high_mired - low_mired ) / 50.0 );
+        double low_mired =
+            std::clamp( std::min( mir1, mir2 ), min_mired, max_mired );
+        double high_mired =
+            std::clamp( std::max( mir1, mir2 ), min_mired, max_mired );
+        double mired_step = std::max( 5.0, ( high_mired - low_mired ) / 50.0 );
 
-    double last_mired = 0.0, estimated_mired = 0.0, current_error = 0.0,
-           last_error = 0.0, smallest_error = 0.0;
+        double last_mired = 0.0, estimated_mired = 0.0, current_error = 0.0,
```
soswow, line 1255: I know you didn't change this for real, just whitespaces, but suggestion is to have some constant that will have value `0.0` but have a name and make it clear that mired value of 0.0 here used as NOT_FOUND basically.

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1357,21 +1475,27 @@ void get_camera_XYZ_matrix_and_white_point(
...
+    {
+        std::vector<double> output_RGB_to_XYZ_matrix =
+            matrix_RGB_to_XYZ( chromaticitiesACES );
+        std::vector<double> output_XYZ_white_point =
+            mulVector( output_RGB_to_XYZ_matrix, deviceWhiteV );
+        CAT_matrix =
+            calculate_CAT( camera_XYZ_white_point, output_XYZ_white_point );
+    }
 
     return CAT_matrix;
```
soswow, line 1498: with this change `CAT_matrix` now can be an empty matrix. HAve we confirmed that all consumers are ready for that as expected?

antond-weta: Will mention in the release notes. Are you suggesting something else?

An empty matrix is better than a broken one. This method has never been used by rawtoaces, and will be marked as deprecated prior to the next release. Also the matrix may be incorrect in some cases according to the discussion above.

soswow: is that a public API? Sorry, I didn't check before commenting. I thought it might be used somewhere inside. In this case - sure, release notes will be enough.

soswow (approved):
lgtm

PR #268, opened 2026-05-04
Fix default orientation and box white balance

remia:
## Description

Proposing the follow minor changes:
* Change the --flip default value from 0 to -1, meaning it will follow the metadata instead of disabling the flip (0 means none), see https://github.com/LibRaw/LibRaw/blob/master/doc/API-datastruct.html#L740
* Pass use_auto_wb = 1 when using box white balancing method, see https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/raw.imageio/rawinput.cpp#L527

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ A help message with a description of all command line options can be obtained by
         --highlight-mode VAL            0 = clip, 1 = unclip, 2 = blend, 3..9 = rebuild. (default: 0)
         --crop-box X Y W H              Apply custom crop. If not present, the default crop is applied, which should match the crop of the in-camera JPEG.
         --crop-mode STR                 Cropping mode. Supported options: 'none' (write out the full sensor area), 'soft' (write out full image, mark the crop as the display window), 'hard' (write out only the crop area). (default: soft)
-        --flip VAL                      If not 0, override the orientation specified in the metadata. 1..8 correspond to EXIF orientation codes (3 = 180 deg, 6 = 90 deg CCW, 8 = 90 deg CW.) (default: 0)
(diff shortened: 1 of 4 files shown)


soswow:
Is there a way we can have a regression tests for changed functionality? Also, this feels like a breaking API  move basically - changing default behaviour.

remia:
I'm not sure there is an easy to test this but will have a look at the test suite.

Also yes this changes behaviour, considering these are bug fix so hopefully this is for the better (the greybox white balance was non functional and the orientation from multiple conversion I tested was wrong, though if some are relying on incorrect orientation metadata in the source file, the latter change might be an issue but for the general case I'd think it's something we want).

antond-weta:
I'm afraid there is no easy way to test orientations. We don't have any vertical images in the test suite, and ideally we would need multiple. Implementing generation of fake raw test images is on the road map, but until then there is not much we can do.

remia:
Ok, I'll leave this here for the time being then, trying to sort out the CLA thing.

remia:
Should be signed now, checking if the update will pick it up.

antond-weta:
Hi @remia, how is it going with CLA? I'd like to push out v2.1.1 with the recent DNG bug fixes, would be great to have this one included as well.

remia:
@antond-weta the CLA was signed last week, unfortunately I don't see any reasons why it fails, "see details" is just a link to the signing page. Wonder if the PR needs to be closed / opened again. Will have a look again tomorrow, sorry about that.

remia:
Looks like adding my email to authorised contributors from the company dashboard website was not enough and I also needed to follow the steps in the EasyCLA link, anyway, looks like it's happy now (I tried to update the commits earlier to bump the date in case that was related so had to force push unfortunately).

antond-weta (approved):

PR #277, opened 2026-05-09
[refactor] avoid Kelvin/mired round-trip in DNG solver (#273)

### LINKED ISSUE #273: [core] Clarify Kelvin vs mired around XYZ_to_color_temperature (avoid redundant round-trips)
## Summary

**`XYZ_to_color_temperature`** interpolates in **mired** space (using `robertson_mired_table`), converts the result to **Kelvin** via `mired_to_kelvin`, clamps, and returns Kelvin. Downstream code often immediately converts that Kelvin value **back to mired** with `kelvin_to_mired` (for example when weighting matrices). That is an extra round-trip that may be unnecessary or lossy from a numerical/design standpoint.

## Background (for newcomers)

Correlated color temperature (CCT) appears both as Kelvin and as **mired** (micro reciprocal Kelvin, \(10^6/\mathrm{K}\)) in this codebase. Helpers live in `src/rawtoaces_core/rawtoaces_core.cpp`: `kelvin_to_mired`, `mired_to_kelvin`, and `XYZ_to_color_temperature`. Some call sites take XYZ → CCT (Kelvin) → mired again within the same workflow.

This issue asks for a **design cleanup** (and tests as appropriate): clarify whether the Kelvin return is the right API, whether some paths should stay in mired until later, or whether the duplication is intentional and should be **documented**. Do **not** change behaviour blindly—justify any change with tests or references.

## Where to look

- `src/rawtoaces_core/rawtoaces_core.cpp` — `XYZ_to_color_temperature` (interpolation over `robertson_mired_table`, then `mired_to_kelvin`, clamp ~2000–50000 K).
- Same file — call sites such as neutral CCT: `XYZ_to_color_temperature` then `kelvin_to_mired` (see search for `neutral_CCT` / `kelvin_to_mired`).
- `src/rawtoaces_core/rawtoaces_core_priv.h` — declarations exposed for unit tests.
- `tests/testDNGIdt.cpp` — existing tests touching `XYZ_to_color_temperature` / kelvin–mired helpers.

- Trace all callers of `XYZ_to_color_temperature` and note Kelvin vs mired expectations.
- Decide: reduce redundant conversions, narrow API (if maintainers agree), or document why Kelvin at the boundary is required.

thc1006:
## Description

`XYZ_to_color_temperature` interpolated the Robertson table in mired, converted the result to Kelvin and clamped it to `[2000, 50000]` K. Its only caller, `find_camera_to_XYZ_matrix`, converted the result straight back to mired (#273).

This PR replaces the private helper with `XYZ_to_mired`, which returns the interpolated value in mired, clamped to the equivalent `[20, 500]` mired range (`1e6 / 50000` and `1e6 / 2000` are exact doubles, so the clamped values are identical to before). Inside the range the solver now sees the interpolated value itself instead of `1e6 / (1e6 / mired)`, which removes a rounding error of about one ulp; the solver golden (`Found illuminant: 5317k.` and the expected matrix) is unchanged.

The Kelvin helper had no other callers and is removed. It was only declared in `rawtoaces_core_priv.h`, so no installed header, binding or documentation changes. `color_temperature_to_XYZ` still takes Kelvin; happy to follow up on that separately if a mired variant is wanted.

Rebased onto the current `main`.

Resolves #273.

## Tests

The three `testIDT_XYZToColorTemperature*` tests in `tests/testDNGIdt.cpp` are replaced with direct mired tests of `XYZ_to_mired`:

- `testIDT_XYZToMired` - interior value, `179.7053420705` mired (the reciprocal of the previous `5564.6648479019` K golden).
- `testIDT_XYZToMired_LowerClamp` - first Robertson entry clamps to `20` mired.
- `testIDT_XYZToMired_UpperClamp` - past the last Robertson entry clamps to `500` mired.

`testIDT_FindCameraToXYZMtx` is untouched and guards the solver output against any drift from removing the round-trip. (The 5317k value it pins turned out to be wrong for an unrelated reason, see #325 and #327; this PR keeps it on purpose, it only removes the round-trip and does not change what the solver computes.)

Verified locally on Ubuntu 24.04 (gcc 13.3, C++17, Eigen + Ceres + lensfun, shared libs): build with `-Werror -Wall -Wextra -pedantic`, all 17 ctest targets, `cmake --install` and `tests/config_tests`. Also clang 18 with `RTA_SANITISER_MODE=address` running `tests/run_valid.sh` (no reports), CodeChecker (clangsa + clang-tidy) and cppcheck on `rawtoaces_core.cpp` (no new findings), and clang-format 16 (no diffs).

## Checklist:

- [x] I have updated the documentation, if applicable. (Private helper only; no documentation changes needed.)

Assisted-by: Claude Code / Claude Opus 5 (Used to rebase onto current `main`, apply the review feedback, rewrite the tests and run the CI steps locally; I reviewed every change before pushing.)

### DIFF
--- a/src/rawtoaces_core/rawtoaces_core.cpp
+++ b/src/rawtoaces_core/rawtoaces_core.cpp
@@ -1044,14 +1044,16 @@ double light_source_to_color_temp( const unsigned short tag )
     return 5500.0;
 }
 
-/// Convert XYZ values to correlated color temperature using Robertson method.
-/// This function estimates the color temperature from XYZ values by interpolating
-/// between known color temperature points in CIE 1960 UCS space. It uses the Robertson
-/// method to find the closest color temperature match based on the UV coordinates.
+/// Compute correlated color temperature from XYZ values, returned in mired.
+/// This is the Robertson UV-distance interpolation that XYZ_to_color_temperature
+/// has always performed internally; exposing it lets callers that already work
+/// in mired (such as find_camera_to_XYZ_matrix) skip the Kelvin conversion.
+/// The result is unclamped here; the [2000, 50000] K clamp is applied by the
+/// XYZ_to_color_temperature wrapper.
 ///
-/// @param XYZ XYZ color values [X, Y, Z]
-/// @return Correlated color temperature in Kelvin
-double XYZ_to_color_temperature( const std::vector<double> &XYZ )
+/// @param XYZ XYZ tristimulus values [X, Y, Z]
+/// @return Color temperature in mired (10^6 / Kelvin), unclamped.
+double XYZ_to_mired( const std::vector<double> &XYZ )
 {
     std::vector<double> uv                  = XYZ_to_uv( XYZ );
     int                 num_robertson_table = countSize( robertson_uvt_table );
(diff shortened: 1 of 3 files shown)


src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1271,9 +1287,10 @@ bool find_camera_to_XYZ_matrix(
                 continue;
             }
 
+            // The loop compares errors in mired space, so use the mired
+            // helper directly and skip the Kelvin round-trip.
             auto neutral_XYZ   = mulVector( camera_to_XYZ_matrix, neutral_RGB );
-            auto neutral_CCT   = XYZ_to_color_temperature( neutral_XYZ );
-            auto neutral_mired = kelvin_to_mired( neutral_CCT );
+            auto neutral_mired = XYZ_to_mired( neutral_XYZ );
```
antond-weta, line 1215: This changes the existing functionality as the `neutral_mired` is now unclamped.

As this place is the only one using the function, I would move the clamp inside `XYZ_to_mired` and get rid of `XYZ_to_color_temperature`, even if that would require modifying the unit tests.

thc1006: Thanks, good catch. The old caller observed the Kelvin clamp after converting the result back to mired, so using the raw interpolation changed the solver semantics.

I've moved the equivalent `[20, 500]` mired clamp into `XYZ_to_mired`, removed the Kelvin helper (it had no other callers), and replaced the Kelvin tests with direct mired tests for the interior value and both clamped ends. The change is rebased onto the current `main`, and the existing DNG solver golden is unchanged.

antond-weta (approved):
LGTM

PR #280, opened 2026-05-14
[fix] get pixels from libraw in camera raw instead of XYZ in DNG mode

antond-weta:
## Description

Change the DNG path to request images from libraw in white-balanced camera raw space instead of XYZ.
This way we can control all colour space conversions

## Tests

I have only one camera shooting DNG natively (BMPCC), so I have only run very limited tests. I've shot the ColorChecker under controlled lights, setting the camera white-balance to match the measured CCT. The new mode appears to be more accurate to me, although the current code looks so broken, I'm not sure if that is specific to my camera.

An example:

Old code:
BMPCC_3600K_aces

New code:
BMPCC_3600K_aces_raw

## Checklist:

  doesn't change the API.)
  nonconforming.


antond-weta:
Hey @remia, could you take a look at this please

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1495,10 +1495,6 @@ bool get_camera_XYZ_matrix_and_white_point(
         return false;
     }
 
-    double exposure_scale = std::pow( 2.0, metadata.baseline_exposure );
-    for ( auto &row: out_camera_to_XYZ_matrix )
-        scaleVector( row, exposure_scale );
-
     if ( metadata.neutral_RGB.size() > 0 )
```
soswow, line 1486: Removing the `2^baseline_exposure` scaling on `out_camera_to_XYZ_matrix` here changes behavior for **every** caller of `get_camera_XYZ_matrix_and_white_point`, including `calculate_DNG_CAT_matrix`, not only the DNG debayer path switched to `raw:ColorSpace` in `image_converter.cpp`. It would help reviewers if the PR description called that out explicitly and justified why baseline exposure should no longer scale the solved camera matrix in those code paths as well.

- Sasha's Pet Robot

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1556,45 +1552,62 @@ std::vector<std::vector<double>> MetadataSolver::calculate_CAT_matrix()
...
-        for ( size_t j = 0; j < 3; j++ )
-        {
-            XYZ_D65_acesrgb[i * 3 + j] = XYZ_D65_acesrgb_3[i][j];
-            CAT[i * 3 + j]             = CAT_matrix[i][j];
-        }
+        // The camera_to_XYZ_matrix expects camera raw values, but the pixels
+        // we get from libraw are white-balanced. Undo the white-balancing as
+        // the first step.
+        std::vector<std::vector<double>> result( 3, std::vector<double>( 3 ) );
+        result[0][0] = metadata.neutral_RGB[0];
```
soswow, line 1588: This path indexes `metadata.neutral_RGB[0..2]` without checking `neutral_RGB.size() >= 3`. If `get_camera_XYZ_matrix_and_white_point` could ever succeed with fewer than three multipliers, this would be out-of-bounds. If that is impossible for DNG/metadata here, a short `assert` or explicit guard would document the contract.

- Sasha's Pet Robot

tests/python/test_image_converter.py
```diff
@@ -141,30 +141,30 @@ def test_configure_metadata(self):
...
-            assert abs(idt[2][2] - 1.013915953697747) < 0.0001
+            assert abs(idt[2][0] - -0.0880997582) < 0.0001
+            assert abs(idt[2][1] - -0.483727073) < 0.0001
+            assert abs(idt[2][2] - 1.5700166672) < 0.0001
 
             assert len(transform_matrix[0]) == 3
-            assert abs(transform_matrix[0][0] - 1.0536466144250152) < 0.0001
-            assert abs(transform_matrix[0][1] - 0.00390441818863832) < 0.0001
-            assert abs(transform_matrix[0][2] - 0.004908450238340354) < 0.0001
+            assert abs(idt[0][0] - 0.68096686026) < 0.0001
```
soswow: These assertions still check row lengths on `transform_matrix` but compare numeric literals against `idt[...]` rather than `transform_matrix[...]`. That duplicates the IDT checks and no longer verifies that `transform_matrix` matches the expected values (or matches `idt`) if those ever diverged. Consider restoring `transform_matrix[i][j]` in the comparisons, or explicitly asserting equality between `idt` and `transform_matrix`.

- Sasha's Pet Robot

src/rawtoaces_core/rawtoaces_core.cpp
soswow: The docstring above still says this function applies baseline exposure compensation to the camera matrix, but that scaling was removed in this PR. Consider updating the comment block so it matches the new behavior (otherwise future readers will assume the matrix is still baseline-scaled).

-- Sasha's Pet Robot

remia:
> Hey @remia, could you take a look at this please

Thanks @antond-weta that fix the white balance issue I saw previously, below is a comparison before / after on one extreme case I saw (DNG from AdobeDNGConveter Camera profile):
before
after

There are 2 things I needed to have this match the DNG SDK more closely (obviously not using the Forward matrix),
* use the Bradford CAT instead of CAT02 as recommended by Adobe,
* apply a scaling to the IDT matrix so that the output ACES white is obtained just when the first camera channel saturates (see DNG SDK dng_color_spec::SetWhiteXY scale factor). An example quick and dirty test I did for this scaling was inserting this code just before applying the final XYZ to ACES2065-& matrix:
```
// Scale matrix so ACES white can just be reached when the first camera channel saturates
std::vector<double> output_RGB_white_point_camera = mulVector(invertVM(mulVector(CAT_matrix, transposeVec(camera_to_XYZ_matrix))), output_XYZ_white_point);
double scale = std::max( output_RGB_white_point_camera[0], std::max( output_RGB_white_point_camera[1], output_RGB_white_point_camera[2] ) );
for ( size_t i = 0; i < result.size(); i++ )
{
    for ( size_t j = 0; j < result[i].size(); j++ )
    {
        result[i][j] *= scale;
    }
}
```
I'm not sure those change are very important, but just FYI.

src/rawtoaces_core/rawtoaces_core.cpp
remia: That's a good point and I did wonder as well, my current understanding is that exposure factor was not having any effect previously as the `out_camera_to_XYZ_matrix` is never used and the `out_camera_XYZ_whitepoint` is normalized at the end of `get_camera_XYZ_matrix_and_white_point`.

antond-weta: I believe the scaling had been put there by mistake and made no difference, as the matrix was only used for white-balance calculation and discarded right after. This PR doesn't make any changes to the exposure adjustments, only the colour shift. 

Interestingly, the DNG spec doesn't contain any instructions on using the `BaselineExposure` and `BaselineExposureOffset`. I'd suggest looking into that separately.

antond-weta:
Let me update this PR to use Bradford, and we'll address the scaling issues (if any) separately.

We should definitely look at using the Forward matrices, and the third calibration illuminant. Not sure how much of that is now available from the latest libraw.

soswow (approved):

soswow:
Since I've made this excersize why not post my pictures too.

Before:
Screenshot 2026-05-17 at 6 54 21 pm (2)

After:
Screenshot 2026-05-17 at 6 54 19 pm (2)

Cursor in oiio's `iv` was in the same spot. Crazy high blue values are not there any more

soswow (approved):

PR #284, opened 2026-05-18
[CI] bump the version of ASWF containers

antond-weta:
## Description

Update the version of ASWF containers on CI:
aswf/ci-rawtoaces:2024.5 -> aswf/ci-rawtoaces:2024.7
aswf/ci-rawtoaces:2025.4 -> aswf/ci-rawtoaces:2025.6
aswf/ci-rawtoaces:2026.2 -> aswf/ci-rawtoaces:2026.4

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -105,15 +105,15 @@ jobs:
         shell: bash
         run: |
           bash build_scripts/${{ inputs.install_deps }}.bash
-              
+
       - name: Dependencies (aswf/ci-rawtoaces)
         if: inputs.vfxyear > 2023
         shell: bash
         run: |
           conan profile detect --force
-          conan install --requires=nlohmann_json/3.12.0 --requires=nanobind/2.9.2 --generator CMakeDeps --generator CMakeToolchain --output-folder=build_deps --build=missing
+          conan install --requires=tsl-robin-map/1.4.0 --generator CMakeDeps --generator CMakeToolchain --output-folder=build_deps --build=missing --settings compiler.cppstd=gnu17
           python -m pip install pytest
-          sudo yum install --setopt=tsflags=nodocs -y perl-Image-ExifTool lensfun-devel
+          sudo yum install --setopt=tsflags=nodocs -y perl-Image-ExifTool
     
       - name: Configure CMake
         run: >
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,13 +43,13 @@ jobs:
           - vfxyear: 2023
             container: aswf/ci-osl:2023-clang15.2
           - vfxyear: 2024
-            container: aswf/ci-rawtoaces:2024.5
+            container: aswf/ci-rawtoaces:2024.7
             toolchain_file: "/__w/rawtoaces/rawtoaces/build_deps/conan_toolchain.cmake"
           - vfxyear: 2025
-            container: aswf/ci-rawtoaces:2025.4
+            container: aswf/ci-rawtoaces:2025.6
             toolchain_file: "/__w/rawtoaces/rawtoaces/build_deps/conan_toolchain.cmake"
           - vfxyear: 2026
-            container: aswf/ci-rawtoaces:2026.2
+            container: aswf/ci-rawtoaces:2026.4
             cxx_standard: 20
             toolchain_file: "/__w/rawtoaces/rawtoaces/build_deps/conan_toolchain.cmake"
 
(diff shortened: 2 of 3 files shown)


antond-weta:
@jfpanisset FYI 
I've updated the rawtoaces containers to the latest version for 2024, 2025, 2026. All pre-installed dependencies seem to work fine. I had to manually install tsl-robin-map, which is a transient dependency of nanobind.

I couldn't get the 2023 to work, the pre-installed version of OpenImageIO doesn't appear to have the raw input plugin enabled. Also I had to manually install a C++17-compatible version of clang to make rawtoaces build on 2023.

jfpanisset:
About tsl-robin-map: the Conan recipe used to build nanobind does specify tsl-robin-map as a dependency / requirement:

https://github.com/AcademySoftwareFoundation/aswf-docker/blob/e41baeec9283d8fe1016d2ed6e98a36a454999a5/packages/conan/recipes/nanobind/conanfile.py#L39

so in theory the ci-rawtoaces containers should have tsl-robinmap if they have nanobind:

```
$ docker run  --rm aswftesting/ci-rawtoaces:2026 ls /usr/local/include/tsl
robin_growth_policy.h
robin_hash.h
robin_map.h
robin_set.h
```

But it looks like the recipe to build Nanobind which I got from the Conan Center Index ends up installing Nanobind in a weird place, i.e. `/usr/local/nanobind/...` which in turn means its Cmake files live in `/usr/local/nanobind/cmake` which are likely not getting found, which in turn means tsl-robin-map is not getting found (since that doesn't have its own CMake files).

That also explains why I had to build rawtoaces with:

```
 tc.cache_variables["RTA_BUILD_PYTHON_BINDINGS"] = False # ASWF FIXME: until we can make nanobind work
 ```
 
since I couldn't get nanobind to work correctly in Conan. I'll update the nanobind recipe to have the package end up in more predictable directories in a future version:

https://github.com/AcademySoftwareFoundation/aswf-docker/issues/399

For the 2023 image, that includes OpenImageIO 2.4.17.0, in theory all versions of OpenImageIO we build are supposed to have libraw enabled, but I've had to do a fair amount of patching to make OIIO's CMake find libraw while building inside Conan, and it's possible that I missed something for the 2023 / 2.4.17.0 build and it ended up getting built without libraw support:

https://github.com/AcademySoftwareFoundation/aswf-docker/issues/400

Finally for the clang / llvm version: the 2023 images include LLVM 14.0.6, which fully supports C++ 17, but C++ 17 only becomes the default with LLVM 16, so you need to specify `-std=c++17` to force building with C++ 17 using LLVM 14:

```
$ docker run -it --rm aswftesting/ci-rawtoaces:2023
[root@1f2234003aa5 aswf]# clang --version
clang version 14.0.6
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/bin
[root@1f2234003aa5 aswf]# cd /tmp
[root@1f2234003aa5 tmp]# cat > toto.cpp
#include <iostream>
#include <map>

int main() {
    std::map<int, std::string> myMap;
    auto [iter, inserted] = myMap.insert({1, "test"});

    std::cout << "Inserted: " << inserted << std::endl;
    return 0;
}
[root@1f2234003aa5 tmp]# /usr/local/bin/clang++ -c toto.cpp
toto.cpp:6:10: warning: decomposition declarations are a C++17 extension [-Wc++17-extensions]
    auto [iter, inserted] = myMap.insert({1, "test"});
         ^~~~~~~~~~~~~~~~
1 warning generated.
[root@1f2234003aa5 tmp]# /usr/local/bin/clang++ -std=c++17 -c toto.cpp
[root@1f2234003aa5 tmp]# 
```

I will address the two issues around nanobind and OIIO libraw, let me know if you are able to get the LLVM 14 included with the 2023 images (whether the older `aswf/ci-osl:2023-clang15.2` you were using or the newer `aswf/ci-rawtoaces:2023.3`.

But is it OK to conclude that you have a working solution for now and I don't need to put out an urgent re-release of the ci-rawtoaces images?

antond-weta:
Thanks JF, 
This is not urgent. 2024-2026 are all working fine now, including the python bindings (albeit with manually installed tsl-robin-map). We can switch the 2023 over later, when it's ready.

From your explanation it seems like c++17 is not an issue on 2023, we just need to enable it. I had to add ` --settings compiler.cppstd=gnu17` to the command line building tsl-robin-map with conan, rawtoaces itself was building fine (although I assumed that it was building successfully because conan had installed the compiler).

soswow (approved):

PR #285, opened 2026-05-18
[refactor] matrix mult and inverse functions

antond-weta:
## Description

Simplify the linear algebra code, remove unused functions.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_core/mathOps.h
+++ b/src/rawtoaces_core/mathOps.h
@@ -5,8 +5,6 @@
 
 #include "define.h"
 
-#include <cfloat>
-
 #include <Eigen/Core>
 #include <ceres/ceres.h>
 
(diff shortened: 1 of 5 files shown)


src/rawtoaces_core/mathOps.h
```diff
@@ -17,56 +15,44 @@ namespace core
...
-/// Calculate the determinant of a square matrix.
-///
-/// @param vMtx a matrix to calculate a determinant for
-/// @return the determinant
-template <typename T> T determinant( const std::vector<std::vector<T>> &vMtx )
-{
-    assert( isSquare( vMtx ) );
+    size_t cols = src_mat[0].size();
+    if ( cols == 0 )
+        return false;
```
soswow: how about 
```suggestion
if (rows != cols) return false;
```

soswow (approved):

PR #289, opened 2026-05-22
[refactor] add rta:core:math namespace

antond-weta:
## Description

Create the `rta::core::math` namespace.
Add helper functions to convert to/from Eigen data structures.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_core/core_math.cpp
+++ b/src/rawtoaces_core/core_math.cpp
@@ -153,7 +153,6 @@ double
 calculate_SSE( const std::vector<double> &tcp, const std::vector<double> &src )
 {
     assert( tcp.size() == src.size() );
-    std::vector<double> tmp( src.size() );
 
     double sum = 0.0;
     for ( size_t i = 0; i < tcp.size(); i++ )
--- a/src/rawtoaces_core/core_math.cpp
+++ b/src/rawtoaces_core/core_math.cpp
@@ -11,6 +11,78 @@ namespace rta
 {
 namespace core
 {
+namespace math
+{
+
+/// Create eigen vector from std::vector
+template <typename T>
+Eigen::Matrix<T, Eigen::Dynamic, 1> vec_from_std( const std::vector<T> &vec )
+{
+    size_t n = vec.size();
+    assert( n > 0 );
+
+    Eigen::Matrix<T, Eigen::Dynamic, 1> result( vec.size() );
+    for ( size_t i = 0; i < n; i++ )
+        result( i ) = vec[i];
+    return result;
+}
+
+/// Create std::vector from eigen vector
+template <typename T>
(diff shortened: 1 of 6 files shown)


src/rawtoaces_core/core_math.cpp
```diff
@@ -11,106 +11,111 @@ namespace rta
...
+        result( i ) = vec[i];
+    return result;
+}
 
-    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> m;
-    m.resize( rows, cols );
-    for ( Eigen::Index i = 0; i < m.rows(); i++ )
-    {
-        if ( src_mat[i].size() != cols )
-            return false;
```
soswow, line 32: 1. Moving methods around makes github show very bad diff.
2. Seems like we've dropped that per-raw validation. There are some asserts, but they are not runtime checks. 
You've removed test 
```
OIIO_CHECK_ASSERT( !inverse( { { 1.0, 2.0 }, { 1.0 } }, result ) );
``` 
that was checking just that.

src/rawtoaces_core/core_math.cpp
soswow: my llm accidentally noticed this not being used. I know it's not in this PR.

src/rawtoaces_core/rawtoaces_core.cpp
```diff
@@ -1422,17 +1424,18 @@ bool get_camera_XYZ_matrix_and_white_point(
...
     {
         out_camera_XYZ_white_point =
-            product( out_camera_to_XYZ_matrix, metadata.neutral_RGB );
+            math::product( out_camera_to_XYZ_matrix, metadata.neutral_RGB );
     }
     else
     {
         out_camera_XYZ_white_point = color_temperature_to_XYZ(
             light_source_to_color_temp( metadata.calibration[0].illuminant ) );
     }
```
soswow, line 1437: Do you want to add some assertion back in?

```suggestion
assert( out_camera_XYZ_white_point.size() == 3 );
assert( out_camera_XYZ_white_point[1] != 0.0 );

```

src/rawtoaces_core/core_math.cpp
antond-weta: I have not moved anything. I inserted 4 new functions on top. Will see if I can work around that.

We can not have runtime checks anymore. I added that to `inverse` in the previous PR, but now this code sits in the separate helper functions, and not all places using it have runtime error checking. I now think the check actually doesn't belong there. All validation should be done upon data ingestion with proper error reporting, not when performing arithmetic operations.

antond-weta: VSCode shows the diff correctly.

soswow (approved):

PR #290, opened 2026-05-23
[refactor] make eigen optional

antond-weta:
## Description

Makes the dependency on Eigen optional via the `RTA_ENABLE _EIGEN` cmake option.
Introduces the `rta::core::math::use_eigen` flag to switch between the built-in and Eigen math. The flag is intended for testing purposes only, we do not have any plans to make it public at this stage.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,17 +65,22 @@ jobs:
       install_deps: install_deps_${{ matrix.os }}
       toolchain_file: ${{ matrix.toolchain_file }}
       build_shared_libs: ${{ matrix.build_shared_libs || 'ON' }}
-      enable_lensfun: ${{ matrix.lensfun }}
-      cxx_standard: 20
-      workflow_name: "lensfun-${{matrix.lensfun}}"
-
+      workflow_name: ${{matrix.options[0]}}
+      enable_eigen: ${{ matrix.options[1] }}
+      enable_lensfun: ${{ matrix.options[2] }}
+      cxx_standard: 17
     strategy:
       # Set fail-fast to false to ensure that feedback is delivered for all matrix combinations. Consider changing this to true when your workflow is stable.
       fail-fast: false
 
       matrix:
         os: [ macos, ubuntu, windows ]
-        lensfun: [ OFF, ON ]
+        options: [
+          # [ name, enable_eigen, enable_lensfun]
+          ["minimal", OFF, OFF],
+          ["eigen",   ON,  OFF],
+          ["full",    ON,  ON ]
+        ]
         include:
           - os: windows
             c_compiler: cl
--- a/configure.cmake
+++ b/configure.cmake
@@ -4,7 +4,10 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_INSTALL_PREFIX}/share/CMake"
 
 find_package ( nlohmann_json CONFIG REQUIRED )
 find_package ( OpenImageIO   CONFIG REQUIRED )
-find_package ( Eigen3        CONFIG REQUIRED )
+
+if ( RTA_ENABLE_EIGEN )
+    find_package ( Eigen3 CONFIG REQUIRED )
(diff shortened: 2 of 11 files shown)


configure.cmake
soswow, line 15: `cmake -S . -B build-review-no-eigen -D RTA_ENABLE_EIGEN=OFF ...`
gives
`Missing required Ceres dependency: Eigen`

soswow:
`config/RAWTOACESConfig.cmake.in` still requires Eigen3 unconditionally. IS that a problem?

src/rawtoaces_core/core_math.cpp
```diff
@@ -84,6 +118,8 @@ mat_to_std( const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> &mat )
     return result;
 }
 
+#endif // RTA_HAS_EIGEN
+
 bool inverse(
```
soswow, line 123: should we rename or do something about this function that will make it clear that it can only do 3x3 matrix, since that's what non-eigen codepath is capable of and what we use it only for.

.github/workflows/ci.yml
```diff
@@ -65,17 +65,22 @@ jobs:
...
       toolchain_file: ${{ matrix.toolchain_file }}
       build_shared_libs: ${{ matrix.build_shared_libs || 'ON' }}
-      enable_lensfun: ${{ matrix.lensfun }}
-      cxx_standard: 20
-      workflow_name: "lensfun-${{matrix.lensfun}}"
-
+      workflow_name: ${{matrix.options[0]}}
+      enable_eigen: ${{ matrix.options[1] }}
+      enable_lensfun: ${{ matrix.options[2] }}
+      cxx_standard: 17
```
soswow, line 71: interesting. we downgrading this here ...

src/rawtoaces_core/define.h
```diff
@@ -43,11 +43,11 @@ namespace rta
 namespace core
 {
 
-const double pi = 3.1416;
+const double k_pi = 3.1416;
```
soswow, line 46: 1. Don't c++ has some sort of convention to call constants (especially these types ones) capitilized?
2. I was looking at the usage sowhere above `if ( tmpXYZ[i][j] > T( k_e ) )` and I was very confused what `k_e` was.  I had to search it and find this line. 

My humble suggestion - can we call it smth like `PI` or if you want `CONSTANT_PI` and `CONSTANT_E` ... wait, what is this even? E is 2.718... 
ok. It's  CIE Lab constants. So, maybe `CIE_LAB_EPSILON` and `CIE_LAB_KAPPA`

soswow (approved):

src/rawtoaces_core/define.h
soswow: considering next lines are 
```
// Planck's constant ([J*s] Joule-seconds)
const double plancks_constant = 6.626176 * 1e-34;
// Boltzmann constant ([J/K] Joules per Kelvin)
const double boltzmann_constant = 1.380662 * 1e-23;
```
pi should prob. be `pi_constant` if we want to follow existing convention

configure.cmake
antond-weta: Can't reproduce. Were you trying to uninstall Eigen? It is still required to be present on the system for Ceres to work.

.github/workflows/ci.yml
antond-weta: no. we officially support c++17 and CI should be running on c++17

src/rawtoaces_core/define.h
antond-weta: k_e is going away soon

configure.cmake
soswow: >  It is still required to be present on the system for Ceres to work.

hm. I might be missing something then. If you can't get rid of Ceres, and Ceres is requireing Eigen, and RTA requires Ceres - how would this flag work?

antond-weta: It makes the Eigen-specific code optional. Same treatment for Ceres is coming in a following PR.

PR #293, opened 2026-05-28
[test] better spectral curves for testing

antond-weta:
## Description

Generate spectral curves mimicking the camera, CMF and, illuminant and training data.
The curves approximate real world curves, and implemented in a form of parametric function, with only 4 numbers stored per curve.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/tests/test_data_helpers.cpp
+++ b/tests/test_data_helpers.cpp
@@ -7,11 +7,12 @@
 #include <fstream>
 
 // Parametric spectral curve in the form of Gaussian function with asymmetric
-// shape: f(x) = a * exp( 1 / ((x - b)/2c)^2 ) )
+// shape: f(x) = a * exp( - (x - b)^2 / 2c^2 ) )
 // Where:
-//    a - amplitude
-//    b - central wavelength
-//    c - shape, c = c1 to the left of the central wavelength, c2 otherwise.
+//    a - maximum value
+//    b - wavelength at peak
+//    c - shape, c = c1 to the left of the peak, c2 otherwise.
+//    HWHM is roughly 1.177 * c
 struct SpectralCurve
 {
     const char *name;
@@ -186,6 +187,7 @@ void save_spectral_json(
         }
         json_data["spectral_data"]["data"][set_name]  = json_spectral_data;
         json_data["spectral_data"]["index"][set_name] = channels;
+        json_data["spectral_data"]["units"]           = "relative";
     }
 
     std::ofstream file( filename );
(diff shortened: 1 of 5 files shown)


tests/test_data_helpers.cpp
```diff
@@ -0,0 +1,193 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Contributors to the rawtoaces Project.
+
+#include "test_data_helpers.h"
+
+#include <nlohmann/json.hpp>
+#include <fstream>
+
```
soswow, line 8: do you want to include #include <cassert> / #include <cmath> ?

tests/test_data_helpers.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+rta::core::SpectralData
+create_hypothetical_camera( const std::string &make, const std::string &model )
+{
+    return generate_spectral_data( HypotheticalCamera, make, model, "" );
+}
+
+rta::core::SpectralData
+create_hypothetical_illuminant( const std::string &type )
```
soswow, line 120: someone seeing this method might think changing type might change the curve shape, but it won't right?

tests/test_data_helpers.cpp
```diff
@@ -0,0 +1,193 @@
...
+{
+    return generate_spectral_data( HypotheticalObserver );
+}
+
+rta::core::SpectralData create_hypothetical_training_data()
+{
+    return generate_spectral_data( HypotheticalTrainingData );
+}
+
+void save_spectral_json(
```
soswow, line 135: do we want to write `units` (= "relative") ?

tests/test_data_helpers.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+#include <nlohmann/json.hpp>
+#include <fstream>
+
+// Parametric spectral curve in the form of Gaussian function with asymmetric
+// shape: f(x) = a * exp( 1 / ((x - b)/2c)^2 ) )
+// Where:
+//    a - amplitude
+//    b - central wavelength
+//    c - shape, c = c1 to the left of the central wavelength, c2 otherwise.
```
soswow: not what you do in the code.
The correct  version would be
```
// Asymmetric Gaussian peak:
//   f(λ) = a * exp( - (λ - b)² / (2 c²) )
//   c = c1 for λ < b, c2 otherwise.
//   a = peak amplitude, b = center wavelength (nm), c = width (nm).
```
if you want to keep the code and update the comment

soswow (approved):

PR #298, opened 2026-06-18
[ci] fix VFX2022 runner

antond-weta:
## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/rawtoaces/blob/main/CONTRIBUTING.md).
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -85,6 +85,8 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: '0'
+        env:
+          ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
 
       - name: Yum config - VFX2022 only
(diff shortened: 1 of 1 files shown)


.github/workflows/build-steps.yml
```diff
@@ -85,6 +85,8 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: '0'
+        env:
+          ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
```
soswow, line 89: what's happening? Should this be a signal to do something with our node version instead?

soswow (approved):

.github/workflows/build-steps.yml
antond-weta: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
that OS is too old to run a more recent version

PR #302, opened 2026-07-04
[refactor] make ceres optional

antond-weta:
## Description

Implements a non-linear solver making the dependency on Ceres optional. Both solvers are available when built with Ceres, and the Ceres is used by default. The solvers are switchable in runtime, but that option is not currently exposed to the users, only being used in unit tests.

## Tests

Yes, full coverage.
There are extensive tests comparing the new solver with the old one.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_core/core_math.cpp
+++ b/src/rawtoaces_core/core_math.cpp
@@ -586,7 +586,10 @@ template <typename T> struct is_same<T, T>
     static const bool value = true;
 };
 
+// (6/29)^3, used in lab_F() and lab_F_prime().
 constexpr double v_6_29__3 = 6.0 * 6.0 * 6.0 / 29.0 / 29.0 / 29.0;
+
+// (29/6)^2, used in lab_F() and lab_F_prime().
 constexpr double v_29_6__2 = 29.0 * 29.0 / 6.0 / 6.0;
 
 // The non-linear perceptual function of the XYZ to LAB conversion.
--- a/src/rawtoaces_core/core_math.cpp
+++ b/src/rawtoaces_core/core_math.cpp
@@ -3,15 +3,24 @@
 
 #include "core_math.h"
 #include "define.h"
+#include <assert.h>
+#include <cmath>
 
 #if defined( RTA_ENABLE_EIGEN ) && RTA_ENABLE_EIGEN
 #    include <Eigen/Core>
+#    include <Eigen/Dense>
+#    include <Eigen/QR>
 #    define RTA_HAS_EIGEN 1
 #else
 #    define RTA_HAS_EIGEN 0
 #endif
 
-#include <ceres/ceres.h>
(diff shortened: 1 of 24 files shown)


src/rawtoaces_core/core_math.cpp
```diff
@@ -490,6 +536,196 @@ getCalcXYZt( const std::vector<std::vector<T>> &RGB, const T beta_params[6] )
...
+    static const bool value = false;
+};
+
+template <typename T> struct is_same<T, T>
+{
+    static const bool value = true;
+};
+
+constexpr double v_6_29__3 = 6.0 * 6.0 * 6.0 / 29.0 / 29.0 / 29.0;
+constexpr double v_29_6__2 = 29.0 * 29.0 / 6.0 / 6.0;
```
soswow, line 593: literally magic numbers, lol
comment would be super helpful here

src/rawtoaces_core/core_math.cpp
```diff
@@ -490,6 +536,196 @@ getCalcXYZt( const std::vector<std::vector<T>> &RGB, const T beta_params[6] )
...
+
+// The non-linear perceptual function of the XYZ to LAB conversion.
+template <typename T> T lab_F( T arg )
```
soswow, line 596: why not call it then `XYZ_to_LAB`?
upd: ah! I can see one below )

src/rawtoaces_core/core_math.cpp
```diff
@@ -527,23 +766,119 @@ struct IDTOptimizationCost
...
+            // Ro = x0*(Rs-Bs) + x1*(Gs-Bs) + Rw*Bs
+            // Go = x2*(Rs-Bs) + x3*(Gs-Bs) + Gw*Bs
+            // Bo = x4*(Rs-Bs) + x5*(Gs-Bs) + Bw*Bs
+
+            // Ro_ = (Rs-Bs) when j == 0, 0 otherwise
+            // Ro_ = (Gs-Bs) when j == 1, 0 otherwise
+            // Go_ = (Rs-Bs) when j == 2, 0 otherwise
+            // Go_ = (Gs-Bs) when j == 3, 0 otherwise
+            // Bo_ = (Rs-Bs) when j == 4, 0 otherwise
+            // Bo_ = (Gs-Bs) when j == 5, 0 otherwise
```
soswow, line 846: what are these above? Forgotted commented out code or part of explaination?

src/rawtoaces_core/core_math.cpp
antond-weta: added

src/rawtoaces_core/core_math.cpp
antond-weta: this explains what is in the matrix out_rgb_ below

soswow (approved):
GOOD

PR #309, opened 2026-07-27
Extract check_input from process_image for focused unit tests

### LINKED ISSUE #274: [ImageConverter] Extract check_input from process_image for focused unit tests
## Summary

Extract the **early validation** at the start of **`ImageConverter::process_image`** (empty path, input existence / filesystem errors) into a dedicated function (working name: **`check_input`**) and invoke it from `process_image`. **Unit tests** should be able to exercise this validation path **directly** instead of relying only on the full `process_image` pipeline (configure, decode, write, etc.).

## Background (for newcomers)

Today, checks live inline at the top of `process_image` in `src/rawtoaces_util/image_converter.cpp` (~empty filename, then `std::filesystem::exists` inside a try/catch). Tests such as `test_last_error_message_empty_filename` and `test_last_error_message_file_not_found` in `tests/test_image_converter.cpp` already call `process_image`, but splitting **`check_input`** makes it easier to test **input validation only** fast and deterministically, and keeps responsibilities clearer for future Dev Days contributors.

## Where to look

- `src/rawtoaces_util/image_converter.cpp` — `ImageConverter::process_image` (opening block through filesystem validation).
- `include/rawtoaces/image_converter.h` — whether `check_input` should be `public` for tests/bindings or test-only via a test helper pattern consistent with this project.
- `tests/test_image_converter.cpp` — tests that assert `status` / `last_error_message` for invalid inputs.
- `tests/python/test_image_converter.py` — Python tests that call `process_image` for invalid paths (behaviour should remain consistent unless deliberately changed with docs).

- Factor the validation block into something like `bool ImageConverter::check_input(const std::string &input_filename)` (or free function if that fits the codebase better).
- Update or add C++ tests that call **`check_input`** where it reduces reliance on the full conversion path.
- Preserve existing observable behaviour for API consumers unless there is a documented reason to change it.

La1itchauhan:
## Description

This PR extracts the early input validation logic from
`ImageConverter::process_image()` into a dedicated
`ImageConverter::check_input()` function.

The validation logic and observable behaviour remain unchanged. This
refactor allows the validation path to be exercised directly in unit
tests without relying on the full image processing pipeline.

Closes #274.

## Tests

Added direct unit tests for `check_input()` covering:

- Empty input filename
- Non-existent input file

The existing `process_image()` validation tests were retained to ensure
the public behaviour remains unchanged.

## Checklist:

- [x] I have updated the documentation, if applicable. (No documentation update required for this refactor.)

### DIFF
--- a/src/rawtoaces_util/image_converter.cpp
+++ b/src/rawtoaces_util/image_converter.cpp
@@ -2627,13 +2627,16 @@ bool ImageConverter::save_image(
     return true;
 }
 
-bool ImageConverter::check_input( const std::string &input_filename )
+bool check_input(
+    const std::string      &input_filename,
+    ImageConverter::Status &status,
+    std::string             &error_message )
 {
     // Early validation: check if input file exists and is valid
     if ( input_filename.empty() )
     {
-        status             = Status::EmptyInputFilename;
-        last_error_message = "Empty input filename provided.";
+        status        = ImageConverter::Status::EmptyInputFilename;
+        error_message = "Empty input filename provided.";
         return false;
     }
 
@@ -2643,16 +2646,16 @@ bool ImageConverter::check_input( const std::string &input_filename )
     {
         if ( !std::filesystem::exists( input_filename ) )
         {
-            status = Status::InputFileNotFound;
-            last_error_message =
+            status = ImageConverter::Status::InputFileNotFound;
+            error_message =
                 "Input file does not exist: '" + input_filename + "'.";
             return false;
         }
     }
     catch ( const std::filesystem::filesystem_error &e )
     {
-        status = Status::FilesystemError;
-        last_error_message =
+        status = ImageConverter::Status::FilesystemError;
+        error_message =
             std::string( "Filesystem error while checking input file '" ) +
             input_filename + "': " + e.what();
         return false;
@@ -2663,7 +2666,7 @@ bool ImageConverter::check_input( const std::string &input_filename )
 
 bool ImageConverter::process_image( const std::string &input_filename )
 {
-    if ( !check_input( input_filename ) )
+    if ( !check_input( input_filename, status, last_error_message ) )
     {
         return false;
     }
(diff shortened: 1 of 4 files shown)


src/rawtoaces_util/image_converter.cpp
```diff
@@ -2627,7 +2627,7 @@ bool ImageConverter::save_image(
     return true;
 }
 
-bool ImageConverter::process_image( const std::string &input_filename )
+bool ImageConverter::check_input( const std::string &input_filename )
```
antond-weta: I would suggest making this a free function for now, not a method of ImageConverter. It doesn't need to be a part of the public interface. Please remove the declaration from the header. If you need the function exposed somewhere for unit testing, declare it in `rawtoaces_util_priv.h` instead

tests/test_image_converter.cpp
```diff
@@ -2462,7 +2462,39 @@ void test_last_error_message_file_not_found()
             std::string::npos,
         true );
 }
+void test_check_input_empty_filename()
```
antond-weta: That functionality has already been covered with unit tests. If you'd like to make changes to unit tests to better reflect the code layout, you may want to edit the existing tests instead of adding new ones.

antond-weta:
Hi @La1itchauhan, thanks for doing this! Looking good, just raised a couple minor things.
Also please check the clang-format check issues above.
Thanks

antond-weta (commented):
Sorry, forgot to submit the review

src/rawtoaces_util/image_converter.cpp
La1itchauhan: Updated — check_input is now a free function declared in rawtoaces_util_priv.h (following the same pattern as fetch_missing_metadata). process_image now passes its own status and last_error_message through by reference, so its external behavior remains unchanged.

tests/test_image_converter.cpp
La1itchauhan: removed the duplicate test functions as you suggested to better reflect the code layout and instead updated the existing test_last_error_message_empty_filename and test_last_error_message_file_not_found tests to call check_input directly. My thinking was that this better addresses #274 by testing the validation logic independently, without introducing additional test cases. If you'd prefer those tests to remain focused on process_image and keep this change more minimal, happy to adjust.

antond-weta:
Looks good! Please address the clang-format issues and I'll get this merged in. Thanks!

PR #315, opened 2026-08-13
[feat] adds --bad-pixels-path

antond-weta:
## Description

adds `--bad-pixels-path` parameter which provides a list of bad pixels to libraw.

## Tests

yes

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/include/rawtoaces/image_converter.h
+++ b/include/rawtoaces/image_converter.h
@@ -212,6 +212,12 @@ public:
         /// 'DHT', 'AAHD', 'AHD'.
         std::string demosaic_algorithm = "AHD";
 
+        /// A path to the file containing a list of bad bixels in libraw format:
+        /// a plain text where each line describes a single bad pixel using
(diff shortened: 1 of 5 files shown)


soswow (approved):
fair enough. lgtm

include/rawtoaces/image_converter.h
```diff
@@ -212,6 +212,12 @@ class ImageConverter
         /// 'DHT', 'AAHD', 'AHD'.
         std::string demosaic_algorithm = "AHD";
 
+        /// A path to the file containing a list of bad bixels in libraw format:
+        /// a plain text where each line describes a single bad pixel using
+        /// three numbers separated by whitespace for the column, row and UNIX
+        /// timestamp.
```
soswow, line 218: what's unix timestamp is for? o_O

antond-weta: to ignore the bad pixels which appeared after the image had been taken

PR #322, opened 2026-09-02
[fix] remove unsupported demosaic algos

### LINKED ISSUE #318: Identical output for --demosaic options AHD AHD-Mod AFD VCD Mixed LMMSE AMaZE
So, I was testing different demosaic options (so far I prefer AAHD, DHT, PPG, in that order quality-wise), but noticed that a large chunk of options produced identical output for me:

AHD
AHD-Mod
AFD
VCD
Mixed
LMMSE
AMaZE

antond-weta:
## Description

A bunch of demosaicing algos have been removed from Libraw some time ago: AHD-Mod, AFD, VCD, Mixed, LMMSE, AMaZE.

Rawtoaces checks if the value provided by the user is one of the supported types and passes it to OIIO.
OIIO parses the string and maps it to numeric `user_qual` param passed to Libraw.
Libraw still accepts the removed user_qual values and falls back to AHD is the requested mode is no longer supported.

With the proposed change rawtoaces will now hard fail if one of the mentioned algos is requested. I'm not certain if this is the best way, as both OIIO and Libraw would still accept the value. Perhaps it would be better to show a warning until this is also addressed on the OIIO side.

## Checklist:

  doesn't change the API.)
  nonconforming.


antond-weta:
@lgritz FYI

soswow (approved):

PR #323, opened 2026-09-05
[fix] make the transform cache thread-safe

### LINKED ISSUE #319: Concurrency issues with caching
I don't know if the aim is to make the library thread safe, but I'm hitting issues in the current caching system.

While processing multiple frames in parallel, I'm noticing that the cache system (I'm actually _disabling_ it but it's still somewhat used) is not thread safe. Caches are static objects where a map is used and entries are returned by reference, when another thread queries an item with the cache setting disabled, the very first step is to [clear](https://github.com/AcademySoftwareFoundation/rawtoaces/blob/main/src/rawtoaces_util/cache_base.h#L76) the map which can result in crash / use-after-free for the thread that was using the last returned entry.

Naive fix I'm using is to protect the fetching function with a mutex and return by value, not the most optimal way but all the cached computations are not my bottleneck by far (also why I disabled it as I hit some issues with bad cache in the past).

When the cache is disabled maybe the caching data structure should not be used.

antond-weta:
## Description

make the transform cache thread-safe

## Tests

yes

## Checklist:

- [x] I have read the [Policy on AI Coding Assistants](https://github.com/AcademySoftwareFoundation/rawtoaces/blob/main/docs/AI_Policy.md)
  and if I used AI coding assistants, I have an `Assisted-by: TOOL / MODEL`
  line in the pull request description above.
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/rawtoaces_util/cache_base.h
+++ b/src/rawtoaces_util/cache_base.h
@@ -63,9 +63,9 @@ template <class Descriptor, class Data> class Cache
 public:
     Cache( const std::string &cache_name = "default" ) : name( cache_name ) {}
 
-    const std::pair<bool, Data> &fetch(
-        const Descriptor                        &descriptor,
-        const std::function<bool( Data &data )> &func )
+    std::shared_ptr<const Data> fetch(
+        const Descriptor                                   &descriptor,
+        const std::function<std::shared_ptr<const Data>()> &func )
     {
         if ( disabled )
         {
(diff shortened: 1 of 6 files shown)


antond-weta:
fixes https://github.com/AcademySoftwareFoundation/rawtoaces/issues/319

soswow (approved):
LGTM.
Being a good meat proxy boy, here is what LLM has to say. Nothing super bad, so feel free to ignore:

## Issues and suggestions

### Critical / should fix

1. **Exception safety in `Cache::fetch`**
   If `func()` throws after a descriptor is added to `_pending_keys`, the pending entry is never removed and waiting threads can deadlock. Consider a RAII guard or `try/catch` around `func()` that always erases the pending key and calls `notify_all()`.

2. **Missing standard includes in `cache_base.h`**
   The header uses `std::mutex`, `std::condition_variable`, and `std::shared_ptr` but does not include `<mutex>`, `<condition_variable>`, or `<memory>`. It compiles today via transitive includes, but that is fragile and may break with different include orders or toolchains.

3. **Missing includes in `testCache.cpp`**
   `test_cache_threading()` uses `std::thread` and `std::vector` without `#include <thread>` or `#include <vector>`. Same fragility as above.

### Medium / worth addressing

4. **Unrelated `nthreads` change in `solve_vignette_map`**
   `nthreads` was changed from `1` to `0` in `solve_vignette_map`. In OIIO, `0` usually means “use all available threads.” That may be intentional, but it is unrelated to the cache/thread-safety fix and changes runtime behavior/performance. Consider reverting or calling it out explicitly in the PR.

5. **Failed results are still cached**
   `nullptr` is inserted into the cache on failure, so later requests return the cached failure without recomputing. This matches the old `{success=false, data}` behavior, but it is worth confirming that is still desired—especially for transient failures.

6. **Possible reentrant deadlock**
   If `func()` internally calls `fetch()` on the same cache with the same descriptor, the inner call will see the key in `_pending_keys`, wait forever, and deadlock. Probably not an issue today, but worth a short comment if reentrancy is not supported.

### Minor / nice to have

7. **`notify_all()` wakes all waiters**
   Correct, but inefficient when many threads wait on different keys. `notify_one()` or per-key wait queues would be more scalable; probably fine for current usage.

8. **Verbosity logging outside the lock**
   “searching for an entry” is logged before acquiring `_mutex`, so multithreaded logs can interleave oddly. Cosmetic only.

9. **Public API signature changes**
   `solve_vignette_map`, `solve_distortion_map`, `solve_aberration_map` now return `std::shared_ptr<const OIIO::ImageBuf>` instead of `bool` + out-parameter. Fine if these are internal/test APIs; mention in the PR if anything external depends on the old signatures.

Issue #326, opened 2026-09-11
exiftool calls fail when popen() leaves errno set (Docker default seccomp)


thc1006:
rawtoaces main at 868a571, Ubuntu 24.04 inside Docker with the default seccomp profile, gcc 13.3, exiftool 12.76 in PATH.

Test_Exiftool and Test_ImageConverter (test_fetch_missing_metadata) fail although exiftool is installed and works from the shell:

```
FAILED: success == should_succeed
	values were 'Failed to execute exiftool. Please make sure that its location is available in PATH. Alternatively you can provide the path to the exiftool binary via the RAWTOACES_EXIFTOOL_PATH environment variable.' and ''
```

`execute()` in src/rawtoaces_util/exiftool.cpp decides the call failed when errno is set after popen():

```cpp
errno = 0;
FILE *file = popen( command.c_str(), "r" );
bool success = ( errno == 0 );
```

popen() only sets errno when it returns NULL; a successful call is free to leave it non-zero. Under Docker's default seccomp profile glibc's posix_spawn tries clone3 first, gets ENOSYS, falls back to clone and runs the command fine, but errno stays at 38:

```c
errno = 0;
FILE *f = popen( "exiftool -ver", "r" );
/* output=12.76, errno after popen = 38 (Function not implemented) */
```

With `--security-opt seccomp=unconfined` errno is 0 and all 17 tests pass. GitHub's runners don't run under that profile, which is why CI never sees it. The comment above the check already says errno is unreliable, and the empty-output workaround was added because of it, but the errno check is still there and wins.

Two smaller things in the same function: if popen() does return NULL, `pclose( NULL )` is called; and the exit status from pclose() is discarded, so a command that prints something and then fails (exiftool exiting with 1, or 127 from the shell) counts as success as long as stdout is not empty.

PR coming up that checks the NULL return and the pclose() status instead of errno.

PR #328, opened 2026-09-11
[fix] stop using errno to detect popen() failures in exiftool (#326)

thc1006:
## Description

`execute()` in exiftool.cpp used `errno != 0` after popen() as the failure signal. popen() only sets errno when it returns NULL, and glibc can leave errno at ENOSYS after a successful call when clone3 is blocked, which is what Docker's default seccomp profile does. Every exiftool call then fails with "Failed to execute exiftool" although the binary runs fine. The reproduction is in #326.

This checks the NULL return instead, and uses the exit status from pclose(): exit code 127 from sh (9009 from cmd.exe) when the binary cannot be found, or whatever exiftool returns when it fails on the file. The empty-output check stays as a safety net, a successful call always prints at least FileName. When popen() returns NULL we no longer call `pclose( NULL )`.

One case where the status is not available: a host process that ignores SIGCHLD (some daemons do that to avoid zombies) makes waitpid fail with ECHILD and pclose() return -1. The old code worked there because it never looked at the status, so for -1 the empty-output check decides on its own, as before. A NULL from popen() takes the same path with no output, so it fails without a separate branch that no test can reach (codecov flagged that line in the first version).

Two smaller things in the same function came out of testing this. The read loop retries on EINTR: a host that installs signal handlers without SA_RESTART (Python does that for any handler set from Python) can get the read interrupted by SIGCHLD when exiftool exits and lose the rest of the output; reproduced with a 100 ms timer signal, both main and the first version of this PR returned false with empty output. The loop reads with `fread()` rather than `fgets()`, which soswow spotted in review: when the signal lands half way through a line, `fgets()` returns NULL and drops the part of the line it has already consumed (glibc keeps it only for `EAGAIN`, musl and FreeBSD never do, Apple's libc returns it), so a retry would resume after the missing prefix. `fread()` returns the bytes it copied. And when exiftool ran but failed on the file (it prints `FileName:` and exits 1) the error was reported as the PATH message; the caller now says `Exiftool failed to read <file>` when there was output and keeps the PATH message when there was none.

`execute()` is now declared in rawtoaces_util_priv.h so the tests can call it directly.

## Tests

- `testExiftool_execute_success`: `echo hello` succeeds and the output is captured.
- `testExiftool_execute_failure_with_output`: `echo hello && exit 3` has to fail. The old code returned success here because stdout was not empty, so this is the one that guards the exit status check.
- `testExiftool_execute_missing_binary`: a command that does not exist fails.
- `testExiftool_execute_sigchld_ignored` (not on Windows): with SIGCHLD set to SIG_IGN, `echo hello` still counts as a success. I checked with a small C program that pclose() does return -1 with ECHILD in that state on glibc.
- `testExiftool_execute_popen_failure` (not on Windows): RLIMIT_NOFILE lowered to 3 so popen() cannot create its pipe and returns NULL; must fail, not crash. On main this segfaults in `pclose( NULL )`.
- `testExiftool_tool_failure_message` (not on Windows): RAWTOACES_EXIFTOOL_PATH points at a script that prints `FileName:` and exits 1; the message has to say exiftool failed on the file.
- `testExiftool_execute_interrupted_read` (not on Windows): a 50 ms interval timer whose handler is installed without SA_RESTART, and a child that prints `Make: NI`, sleeps one second and prints `KON CORPORATION`. The output has to come back complete; with the `fgets()` loop of the first version it comes back as `KON CORPORATION`.

Rebased onto the current `main` after #330 landed.

The seccomp case itself cannot be reproduced on the GitHub runners (no seccomp there). I verified it in Docker: with the default profile Test_Exiftool and Test_ImageConverter fail on main and pass with this change, 17/17 overall; with `seccomp=unconfined` both before and after pass. The commands used in the tests are plain `echo` / `exit`, which work in sh and in cmd.exe, so the Windows runners should be fine too.

Ubuntu 24.04 with gcc 13.3: full ctest, install and config_tests pass under both seccomp profiles, as root and as an unprivileged user; ASan+UBSan and TSan with clang 18 on Test_Exiftool and tests/run_valid.sh clean; CodeChecker and clang-format 16 clean. Also checked outside the suite: output longer than the 255 byte read buffer and 8 MB of output arrive intact, 3000 sequential calls leak no file descriptors, and 8 threads calling `execute()` at once do not mix output.

## Checklist:

- [x] I have updated the documentation, if applicable. (Internal function, no documentation change needed.)

(Used to track down the errno behaviour with a small popen() probe and to run the CI steps locally, including the seccomp comparison. For the fgets() review it compared the fgets() sources of glibc, musl, FreeBSD and Apple's libc, drafted the interrupted-read test and ran the CI steps and the test-reverting check locally; I reviewed the change myself.)


src/rawtoaces_util/exiftool.cpp
```diff
@@ -77,29 +76,55 @@ bool execute( const std::string &command, std::stringstream &stream )
...
+            {
+                stream << buffer;
+                empty = false;
+            }
+            // A signal delivered to the host process (SIGCHLD when the child
+            // exits, for one) interrupts the read when its handler was
+            // installed without SA_RESTART, as Python does. Keep reading.
+            else if ( ferror( file ) && errno == EINTR )
+            {
+                clearerr( file );
```
soswow: Clearing the error and retrying fgets() does not preserve a partially consumed line.
On glibc, fgets() can consume bytes before an interrupted underlying read, then return NULL. For example, if the child writes Make: NI, a signal interrupts the read, and the child subsequently writes KON `CORPORATION\n`, the retry loses the prefix. execute() can nevertheless return success, leaving the caller with missing metadata. This conclusion follows from [glibc’s implementation](https://raw.githubusercontent.com/bminor/glibc/master/libio/iofgets.c); I could not reproduce it on Linux though.
Suggested fix: use a byte-counted read loop that preserves every successful read and retries EINTR. Add a Linux regression test that interrupts a read midway through a line and checks the complete output.

soswow (commented):
LGTM now. thank you

Issue #329, opened 2026-09-11
exiftool: paths with spaces or shell characters are passed to the shell unquoted


thc1006:
rawtoaces main at 868a571, Ubuntu 24.04, gcc 13.3, exiftool 12.76.

`perform_exiftool_call` (src/rawtoaces_util/exiftool.cpp) builds the exiftool command line by string concatenation and hands it to popen(), with neither the exiftool path nor the image path quoted:

```cpp
std::string command = exiftool_path + " -S";
...
command += " " + image_path;
```

So a space or a shell character in either path breaks the call. I tried four cases with BatteryPark.NEF copied around (small program calling `fetch_metadata` with `cameraMake` / `cameraModel`):

```
image path without space                 -> OK   make='NIKON CORPORATION'
image path with spaces                   -> OK   make=''
exiftool path with a space               -> FAIL  Failed to execute exiftool. Please make sure that its locati
image name with $ and ()                 -> FAIL  Failed to execute exiftool. Please make sure that its locati
```

with exiftool complaining on stderr for the second one:

```
Error: File not found - /tmp/dir
Error: File not found - with
Error: File not found - space/Battery
Error: File not found - Park.NEF
```

The second case is the bad one: the shell splits the path into four file names, exiftool prints "File not found" for each, but the call still counts as successful because stdout is not empty (`FileName` is printed for the fragments), so `fetch_metadata` returns true with no metadata and the lens correction later reports the camera make as missing. Nothing points at the space in the path.

On Windows the same applies to `C:\Program Files\...` for the binary, and `(1)` in a file name is what every browser download produces.

I'll send a PR that quotes both paths (single quotes for sh, double quotes for cmd.exe with the extra outer pair cmd needs) and adds a test with a `dir with spaces/Battery Park (1).NEF` copy of the fixture.

PR #330, opened 2026-09-11
[fix] quote the paths on the exiftool command line (#329)

thc1006:
## Description

`perform_exiftool_call` concatenates the exiftool path and the image path into the command line unquoted, so a space or a shell character in either path breaks the call. With a space in the image path the call even looks successful, exiftool prints `File not found` for each fragment but stdout is not empty, and `fetch_metadata` returns true with no metadata. The cases I tried are in #329.

Since the line goes through a shell, an unquoted name can also run commands: a file called `a;id;.NEF` runs `id`. Quoting closes that as well.

This quotes both paths. On sh that is single quotes, nothing inside them is interpreted, an embedded single quote becomes `'\''`. On Windows it is double quotes, and because cmd.exe strips the first and the last quote of the line when there are more than two, the whole line gets an extra pair around it (the usual `cmd /c ""prog" "arg""` idiom). The image path is also preceded by `--`, otherwise a relative name starting with `-` is parsed by exiftool as an option (`Invalid TAG name`). I can only test the sh side here, the Windows runners exercise the cmd.exe side through the existing `testExiftool_tool_in_env` / `tool_in_path` tests (relative exiftool path, no spaces) and the new tests.

One thing quoting does not cover on Windows: cmd.exe expands `%VAR%` even inside double quotes, so a name like `100%TEMP%.NEF` would still be mangled there. That needs CreateProcess instead of `_popen` and I left it out. The outer pair of quotes is added in `perform_exiftool_call` rather than in `execute()` to stay out of the lines #328 changes; it can move there once one of the two has landed.

## Tests

- `testExiftool_path_with_spaces`: copies BatteryPark.NEF to `<tmp>/dir with spaces/Battery's Park (1).NEF` and fetches `cameraMake` / `cameraModel` from there; the apostrophe goes through the `'\''` escape on sh. Fails on main (the apostrophe gives a shell syntax error; with spaces alone the call succeeds with an empty make), passes with this change.
- `testExiftool_path_leading_dash`: copies the fixture to `-Battery Park.NEF` in the working directory and fetches `cameraMake` through that relative name. Fails on main, passes with `--`.

Full ctest 17/17, install and config_tests on Ubuntu 24.04 with gcc 13.3; ASan+UBSan and clang-format 16 clean. Outside the test suite I ran 144 generated names through `fetch_metadata` (every printable ASCII character, `'''`, `$'..'`, UTF-8, ~3000 character paths, glob patterns, an exiftool binary reached through a directory with odd characters): on main 104 worked, 7 returned success with empty metadata and 33 failed; with this change all 143 legitimate ones work and the one deliberate command injection attempt is rejected.

This touches exiftool.cpp and testExiftool.cpp like #328 does, in different places; I checked that the two branches merge cleanly in either order.

## Checklist:

- [x] I have updated the documentation, if applicable. (Internal function, no documentation change needed.)

(Used to write the probe that tried the four path variants and to run the CI steps locally; I reviewed the change myself.)

### DIFF
--- a/src/rawtoaces_util/exiftool.cpp
+++ b/src/rawtoaces_util/exiftool.cpp
@@ -102,6 +102,27 @@ bool execute( const std::string &command, std::stringstream &stream )
     return success && !empty;
 }
 
+/// Quote a path for the shell that popen() runs the command with, so
+/// spaces and characters like ( ) or $ in it are passed through as is.
+static std::string quote_path( const std::string &path )
+{
+#if defined( WIN32 ) || defined( WIN64 )
+    return "\"" + path + "\"";
+#else
+    // Single quotes take everything literally. An embedded single quote is
+    // written as '\'' (close, escaped quote, reopen).
+    std::string result = "'";
+    for ( char c: path )
+    {
+        if ( c == '\'' )
+            result += "'\\''";
+        else
+            result += c;
+    }
+    return result + "'";
+#endif
+}
+
 bool perform_exiftool_call(
     const std::string                  &exiftool_path,
     const std::string                  &image_path,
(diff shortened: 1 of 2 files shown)


antond-weta (approved):
LGTM

